Issue #22, opened 2018-01-11
Schedule SelectClock with main clock


turion:
In `FRP.Rhine.Clock.Select`, there is a schedule for two `SelectClock`s of the same main clock, `schedSelectClocks`. There should be a schedule for a `SelectClock` and its main clock.

SolviQorda:
I'm trying to solve this. I'm a beginner and still wrapping my head around Rhine's structure and use of Arrows, but I think I'm getting somewhere. I hope that's ok!

Can I just ask to check I've understood, when you say :
'there needs to be a schedule for `SelectClock' and its main clock',

am I right in reading that as:
the new function need to output a timestamp and an Either value which specifies whether the subclock or the mainClock has ticked?

turion:
You're very welcome to submit a pull request! If there are problems with it, I'll help you with getting the arrow details etc. right.

Yes, you've understood the issue precisely correct. I think maybe we need a `Monoid` instance for the main clock.

Don't worry if the CI fails, that's because of travis issues I'm still working around.

SolviQorda:
Amazing thanks! I'll have a go and try and make a PR soon :+1:

SolviQorda:
When you say 'we maybe need a monoid instance' - do you think that's a separate issue or do you think it's necessary for this function? Would I need to declare that with the other Clock instances in Clock.hs? 

Thank you for any clarification!

SolviQorda:
Hey, so this is where I'm up to so far - is this the right type signature? Also, please could you help me resolve the following error?

```
schedSelectClockAndMain
  :: (Monad m, Monoid cl, Clock m cl)
  => Schedule m cl (SelectClock cl a)
schedSelectClockAndMain = Schedule {..} -- 85
  where
    startSchedule mainClock1 subClock = do
      (runningClock, initialTime) <- startClock
        $ mainClock1 `mappend` mainClock subClock
      let
        runningSelectClock = concatS $ proc _ -> do
          (time, tag) <- runningClock -< ()
          returnA                     -< catMaybes
            [ (time, ) . Left  <$> select mainClock1 tag
            , (time, ) . Right <$> select subClock tag ]
      return (runningSelectClock, initialTime)

```



```
/home/solvi/Development/osprojects/rhineenv/rhine/rhine/src/FRP/Rhine/Clock/Select.hs:85:27: error:
    • Couldn't match type ‘cl’
                     with ‘SelectClock (SelectClock cl0 (Tag cl0)) (Tag cl)’
[...]

PR #56, opened 2018-05-17
added scheduleSelectClockAndMain function

SolviQorda:
As per issue #22


turion:
Thanks, looking good!

PR #61, opened 2018-05-24
Dev hoist rhines

turion:
This is not complete yet, and more commits should be added:

* Parallel Rhines should also be hoisted
* Ideally, Rhines of arbitrary clock trees should be hoisted, but this seems hard

### DIFF
--- a/rhine/src/FRP/Rhine/Schedule.hs
+++ b/rhine/src/FRP/Rhine/Schedule.hs
@@ -99,6 +99,8 @@ rescaledScheduleS Schedule {..} = Schedule startSchedule'
 
 
 -- TODO What's the most general way we can lift a schedule this way?
+-- | Lifts a schedule into the 'ReaderT' transformer,
+--   supplying the same environment to its scheduled clocks.
 readerSchedule
   :: ( Monad m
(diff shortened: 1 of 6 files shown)


rhine/src/FRP/Rhine/Schedule.hs
```diff
@@ -199,3 +239,19 @@ parClockTagInclusion :: ParClockInclusion clS cl -> Tag clS -> Tag cl
...
+
+{-
+-- * Hoist composite clocks along monad morphisms
+
+type family HoistedClock m m' cl where
+  HoistedClock m m' (SequentialClock m cl1 cl2) = SequentialClock m' (HoistedClock m m' cl1)  (HoistedClock m m' cl2)
+  HoistedClock m m' (ParallelClock   m cl1 cl2) = ParallelClock   m' (HoistedClock m m' cl1)  (HoistedClock m m' cl2)
+  HoistedClock m m' cl = HoistClock m m' cl
+
+hoistedClock :: (forall a . m a -> m' a) -> cl -> HoistedClock m m' cl
```
turion: This won't be able to type-check.

turion:
Stuck on #63.

PR #62, opened 2018-05-24
Feature hoistedclockschedule

turion:


turion:
Like #61 this is incomplete because arbitrary clock trees can't be hoisted that way.

turion:
Basically, stuck on #63.

Issue #42, opened 2018-05-09
schedSelectClock should only use a Semigroup


turion:
With GHC 8.4, `Semigroup`s become more of a thing, so it's time to relax the type signature of `schedSelectClock` to requiring `cl` only to be a `Semigroup`.

turion:
We support GHC >= 8, so I think no conditional imports are needed.

SolviQorda:
So I had a go at this - and the following compiles. Have I understood what you were asking for in the issue?

```
schedSelectClocks
  :: (Monad m, Semigroup cl, Clock m cl)
  => Schedule m (SelectClock cl a) (SelectClock cl b)
schedSelectClocks = Schedule {..}
  where
    startSchedule subClock1 subClock2 = do
      (runningClock, initialTime) <- startClock
        $ mainClock subClock1 <> mainClock subClock2
      let
        runningSelectClocks = concatS $ proc _ -> do
          (time, tag) <- runningClock -< ()
          returnA                     -< catMaybes
            [ (time, ) . Left  <$> select subClock1 tag
            , (time, ) . Right <$> select subClock2 tag ]
      return (runningSelectClocks, initialTime)
```

turion:
Yes, that's the idea! Feel free to apply this to all other schedules with a `Monoid` requirement and send a PR. (Use `git grep Monoid`)

SolviQorda:
I've managed to get all but one of the Schedules onto Semigroups and compile.

My only problem is in `Concurrently.hs`, with `concurrentlyWriter`. I'm trying to find a way for Semigroups to work with Monad Transformers. 

I'm happy to keep trying with `concurrentlyWriter`, or to send a PR with all the changes except `concurrentlyWriter` --> whatever is more useful!

turion:
Ah, the `Monoid` in `concurrentlyWriter` is of a different nature! Have a look at [The `Monad` instance for `WriterT`](http://hackage.haskell.org/package/transformers-0.5.5.0/docs/Control-Monad-Trans-Writer-Lazy.html#t:WriterT). It requires `Monoid w`, and that's the sole reason for requiring it in `concurrentlyWriter`.

turion:
So please do send the PR with everything you have so far :)

SolviQorda:
Ah I thought that might be the case - glad that I didn't spend hours trying to refactor something I couldn't :joy:

PR #76, opened 2018-05-29
refactor schedules that use Monoid to use Semigroup instead

SolviQorda:
as per issue #42 - hope that this is ok!

### DIFF
--- a/rhine/src/FRP/Rhine/Clock/Select.hs
+++ b/rhine/src/FRP/Rhine/Clock/Select.hs
@@ -14,6 +14,7 @@ import Data.MonadicStreamFunction.Async (concatS)
 
 -- base
 import Data.Maybe (catMaybes, maybeToList)
+import Data.Semigroup
 
 -- | A clock that selects certain subevents of type 'a',
 --   from the tag of a main clock.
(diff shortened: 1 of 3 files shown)


rhine/src/FRP/Rhine/Clock/Select.hs
```diff
@@ -38,15 +39,15 @@ instance (Monad m, Clock m cl) => Clock m (SelectClock cl a) where
 
 
 -- | A universal schedule for two subclocks of the same main clock.
---   The main clock must be a monoid (e.g. a singleton).
+--   The main clock must be a Semigroup (e.g. a singleton).
```
turion: Ideally, make a haddock link, i.e. `'Semigroup'`

SolviQorda: Cool ok! Should I do the PR again, or is this just for future reference?

turion: For the real experience, push one commit to your fork that fixes it!

SolviQorda: haha ok :)

turion:
Thanks!

PR #84, opened 2018-06-15
refactored count as a type synonym of FixedStep

SolviQorda:
Hope this is useful!


SolviQorda:
Thanks for being patient with me! I've updated the cabal file now :)

PR #94, opened 2018-06-28
type alias simplifications

ggreif:


turion:
I thought this gives problems with not completely applied type variables..?

turion:
I'm not so sure right now whether this can't lead to problems anywhere. For sure the following is problematic:

```haskell
type Foo a b = (a, b)

instance Functor (Foo a) where
   ...
```
But I don't know whether the opposite is ever problematic, which is what you're proposing. Any idea?

ggreif:
Hmmm, you are right, and travis does confirm it :-(

This would work when the original definition was a `data` or `newtype`. But not for `type`, which cannot be partially applied. My bad, sorry for the noise.

PR #95, opened 2018-06-29
Eliminate warnings w.r.t. ghc-8.4.3

ggreif:
This eliminates all the warnings in `rhine/rhine` for the 3 GHC versions used in Travis.

### DIFF
--- a/rhine/rhine.cabal
+++ b/rhine/rhine.cabal
@@ -107,6 +107,9 @@ library
   hs-source-dirs:      src
 
   ghc-options:         -Wall
+                       -Wno-unused-imports
+                       -Wno-unticked-promoted-constructors
+                       -Wno-type-defaults
 
   -- Base language which the package is written in.
   default-language:    Haskell2010
--- a/rhine/src/FRP/Rhine/ResamplingBuffer/MSF.hs
+++ b/rhine/src/FRP/Rhine/ResamplingBuffer/MSF.hs
@@ -30,7 +30,7 @@ msfBuffer = msfBuffer' []
       -> ResamplingBuffer m cl1 cl2 a b
     msfBuffer' as msf = ResamplingBuffer {..}
       where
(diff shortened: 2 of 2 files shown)


rhine/rhine.cabal
```diff
@@ -106,7 +106,8 @@ library
   -- Directories containing source files.
   hs-source-dirs:      src
 
-  ghc-options:         -Wall
+  ghc-options:         -Wall -Wno-unused-imports -Wno-unticked-promoted-constructors
```
turion: Is it possible to put them on separate lines? (Makes it easier for git to merge if there are edits in different places in this section)

turion:
I'm undecided about unticked promoted constructors. Is there a good reason not to use them? Why does GHC warn about them? I find `'MyConstructor` slightly ugly, but if there is a good reason to use it, I will.

rhine/rhine.cabal
ggreif: Done.

ggreif:
I doubt there is a good reason, sounds like a pedantic thing to me.

turion:
I used ticks in some places like https://github.com/turion/rhine/blob/63fc173d93caea52cd2ec8d969f83ed3fbce433c/rhine/src/FRP/Rhine/Clock/Periodic.hs#L68. Maybe I'll revert that globally to have no ticks anywhere. (Note to self: Ask on IRC what the reasoning is)

rhine/rhine.cabal
```diff
@@ -107,6 +107,9 @@ library
   hs-source-dirs:      src
 
   ghc-options:         -Wall
+	                     -Wno-unused-imports
```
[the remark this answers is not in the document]
ggreif: Hmm, there was a commit that had tabs, but I committed (with `--amend`) and force-pushed an update since then.

turion:
Thanks!

PR #99, opened 2018-08-12
Added GlossSimWithEventClock

turion:
With this clock, it is possible to tick the simulation when an event
occurs.


turion:
Some aspects of rhine-gloss are being rewritten. This needs to be merged first

PR #109, opened 2018-09-26
Update to LTS 12.10

ggreif:


turion:
Thanks! rhine-gloss also needs to be updated, and the examples as well.

ggreif:
Would making those (git) softlinks make sense?

turion:
You mean the other `stack.yaml`s should link to the main one? That wouldn't work, since `rhine-gloss` and `rhine-examples` have an extra local dependency on `rhine`.

ggreif:
Ah, yes, I was looking at the `git grep lts-` output, which appeared equal :-(

ggreif:
I was not done yet!!! :-P

ggreif:
I guess I fat-fingered the close button, and github said that the branch is already merged...

Anyway, the commits are in. Another thought: Does travis support ghc-8.6 yet?

turion:
It's hard to say what exactly travis supports. (Carla Iriberri is on it: https://github.com/travis-ci/docs-travis-ci-com/issues/2021) Also note that I haven't added stack to travis. Have you built all three projects with stack?

ggreif:
I have built `rhine` and `rhine-examples` with this change. `rhine-gloss` assumes `nix-shell` which is not available in my setup.

I think this is good to go.

turion:
Thanks a lot, but unfortunately #111 forced me to (temporarily) set the resolver to nightly :/ I'll revert to LTS 13 once it's out.

ggreif:
I have a `free` bump in my fork, but was hesitating to request a pull, as I was not aware of your nightly plans :-)

turion:
Always feel free to bump because of version bounds :) I like to set them tightly (only one major version if possible), so I sometimes need to adjust them.

PR #118, opened 2018-10-13
Effected GHC 8.6.1 compatibility

0xCM:
* This was primarily an exercise in replacing Monad constraints with MonadFail constraints.
* Applied similar changes to Dunai


0xCM:
It looks like the integration build is failing at this point because the build is being conducted via cabal rather than stack : /

turion:
Thanks a lot! But as I said on dunai, I rather want to have it work without `MonadFail`, and https://github.com/ivanperez-keera/dunai/pull/116 is underway to solve that.

As for cabal vs. stack, I agree that stack is the better build system (and my preferred one), but many people still use cabal, so it should build with cabal _and_ stack.

turion:
This is resolved by #119. Thanks anyways!

PR #127, opened 2019-11-19
Fixed typo

walseb:
I was kinda confused by this since there doesn't seem to be a `un-SN` type. But maybe it's correct though, I haven't fully understood this library yet. Oh and thanks for creating this library!

### DIFF
--- a/rhine/src/FRP/Rhine/Type.hs
+++ b/rhine/src/FRP/Rhine/Type.hs
@@ -8,7 +8,7 @@ module FRP.Rhine.Type where
 import FRP.Rhine.SN
 
 {- |
-A 'Rhine' consists of un 'SN' together with a clock of matching type 'cl'.
+A 'Rhine' consists of a 'SN' together with a clock of matching type 'cl'.
 It is a reactive program, possibly with open inputs and outputs.
 If the input and output types 'a' and 'b' are both '()',
 that is, the 'Rhine' is "closed",


turion:
Thanks :) do feel free to ask questions! Did you read the article, was it any help?

walseb:
Ok cool! Yeah it's great especially as kind of a dictionary since the haddocks are rough if you don't know what the building blocks are (like CLSFs, time domains, etc). But it's still a lot to take in compared to yampa. Right now I'm just going through the haddocks and trying to understand how the operators work

turion:
It's good to know what about the library is hard to understand. I guess what's lacking is a long tutorial, but I couldn't find the time for that amongst all my other projects. You can also have a look at https://github.com/turion/sonnendemo, that might give you a flavour of how to use what operator.

walseb:
I think the largest problem for me is just the sheer amount of operators and functions rhine introduces and understanding how they work. I think a cheat sheet where each operator/function is described with a sentence or so would be very useful. Maybe I will try to write one once I get a better understanding. Thanks I will look into that!

turion:
Oh a cheat sheet is an excellent idea. I didn't think about that. I made one today:

https://github.com/turion/rhine/blob/master/CHEATSHEET.md

walseb:
Oh wow nice! It seems only to exist on the master branch right now though

Issue #131, opened 2020-03-29
update to build with dunai-0.6.0


smunix:

turion:
@smunix thanks! Do you want to create a pull request?

smunix:
Here you go, #132

smunix:
> @smunix thanks! Do you want to create a pull request?

Here you go, #132

PR #132, opened 2020-03-29
build against dunai-0.6.0, bump version #131

smunix:


rhine/dunai.nix
```diff
@@ -0,0 +1,15 @@
+{ mkDerivation, base, MonadRandom, simple-affine-space, stdenv
```
turion: Shouldn't we be able to use `dunai` from `stackage`/`nixpkgs`?

rhine/rhine.release.nix
```diff
@@ -0,0 +1,17 @@
+let
+  config = {
+    packageOverrides = nixpkgs: rec {
+      haskellPackages = nixpkgs.haskellPackages.override {
+        overrides = hnew: hold: rec {
+          rhine = hnew.callPackage ./rhine.nix {};
```
turion: Can we also use `callCabal2Nix` here?

rhine/src/FRP/Rhine/ClSF/Util.hs
```diff
@@ -221,9 +223,9 @@ weightedAverageFrom v0 = feedback v0 $ proc ((v, weight), vAvg) -> do
 --   all features below a given time constant @t@.
 --   (Equivalently, it filters out frequencies above @1 / (2 * pi * t)@.)
 averageFrom
-  :: ( Monad m, VectorSpace v
-     , Floating (Groundfield v)
-     , Groundfield v ~ Diff td)
+  :: ( Monad m, VectorSpace v s
+     , Floating (s)
```
turion: I think you can elide these brackets here (and in some other similar places)

rhine/src/FRP/Rhine/Clock/Periodic.hs
```diff
@@ -45,7 +45,7 @@ instance (Monad m, NonemptyNatList v)
   type Time (Periodic v) = Integer
   type Tag  (Periodic v) = ()
   initClock cl = return
-    ( cycleS (theList cl) >>> withSideEffect wait >>> sumS &&& arr (const ())
+    ( cycleS (theList cl) >>> withSideEffect wait >>> (accumulateWith (+) 0) &&& arr (const ())
```
turion: `sumS` still exists in `dunai` afaikt, you have to `import Data.MonadicStreamFunction.Util`.

rhine-examples/rhine-examples.nix
```diff
@@ -0,0 +1,10 @@
+{ mkDerivation, base, random, rhine, stdenv, transformers }:
```
turion: Same here. Does `callCabal2Nix` work as well?

rhine/src/FRP/Rhine/Clock/Periodic.hs
smunix: True, but the new sumS implementation requires instances of The VectorSpace. Integer isn't a VectorSpace.

turion: Fair enough.

smunix:
> In particular, thank you for the `nix` rig! Is it possible to add it to `.travis.yml`? I'm not sure one can CI both cabal and nix on travis.
> 
Sure, let me take a look.

smunix:
how do you reopen this PR ? I closed it by mistake and I can't add more updates to it.

turion:
There you go :)

turion:
I just realised this is ready to go! Thanks a lot @smunix !

PR #140, opened 2020-05-22
Dev clock erasure 2

### LINKED ISSUE #116: Clock erasure
This is a bigger issue that currently blocks a few other issues which will progress much quicker once this one is solved.

I'm working on a general way how to _erase clocks_ from a Rhine, effectively transforming it to a monadic stream function. This `MSF` can then be run by `dunai`'s `reactimate.

## The plan

* Clock values are effectively `MSF m () TimeInfo`.
* Closed `ClSF`s are effectively `MSF (ReaderT TimeInfo m) () ()`, which is isomorphic to `MSF m TimeInfo ()`.
* Composing them yields a closed `MSF m () ()`, so here we can run the composite of clock and `ClSF` as a mere `MSF`, the clock is erased.
* `ResamplingBuffer`s complicate the issue, but only slightly. Any `ResamplingBuffer m a b` (dropping the clock information for now) can be transformed into an `MSF m (Either a ()) (Either () b)` with the additional property that `Left` input yields `Left` output, and the same for `Right`. The `Tag` from the `SequentialClock` will eventually decide which path is taken.
* Composing everything together will give a big `MSF m () ()` that can simply be `reactimated`.

## Advantages

* Possibly: Performance might get better.
* Several issues like #64 will be easier

## Challenges

I'm working on this, but:

* The type level clock information makes the implementation hard.
* I'd like to solve #115 first

turion:
Fixes #116.

* [x] Changelog
* [ ] Bump version
* [x] Merge #139 first
* [x] Documentation


turion:
Instead of bumping versions it's easier to just merge and do the bumping whenever I release.

PR #156, opened 2021-06-11
Add information regarding the behaviour of threadDelay to the documen…

miguel-negrao:
…tation of the Millisecond clock.

### DIFF
--- a/rhine/src/FRP/Rhine/Clock/Realtime/Millisecond.hs
+++ b/rhine/src/FRP/Rhine/Clock/Realtime/Millisecond.hs
@@ -53,6 +53,17 @@ instance GetClockProxy (Millisecond n)
 --   and waits for the remaining time until the next tick.
 --   If the next tick should already have occurred,
 --   the tag is set to 'False', representing a failed real time attempt.
+
+--   Note that this clock internally uses 'threadDelay' which can block
+--   for quite a lot longer than the requested time, which can cause
+--   the clock to miss one or more ticks when using low values of 'n'. 
+--   When using 'threadDelay', the difference between the real wait time 
+--   and the requested wait time will be larger when using 
+--   the '-threaded' ghc option (around 800 microseconds) than when not using
+--   this option (around 100 microseconds). For low values of @n@ it is recommended
+--   that '-threaded' not be used in order to miss less ticks. The clock will adjust 
+--   the wait time, up to no wait time at all, to catch up when a tick is missed.
+
 waitClock :: KnownNat n => Millisecond n
 waitClock = Millisecond $ RescaledClockS FixedStep $ \_ -> do
   initTime <- getCurrentTime


rhine/src/FRP/Rhine/Clock/Realtime/Millisecond.hs
```diff
@@ -53,6 +53,17 @@ instance GetClockProxy (Millisecond n)
 --   and waits for the remaining time until the next tick.
 --   If the next tick should already have occurred,
 --   the tag is set to 'False', representing a failed real time attempt.
+--   Note that this clock internally uses 'threadDelay' which can block
+--   for quite a lot longer than the requested time, which can cause
+--   the clock to miss one or more ticks when using low values of 'n'. 
+--   When using threadDelay, the different between the real wait time
```
ggreif, line 59:
```suggestion
--   When using threadDelay, the difference between the real wait time 
```

miguel-negrao:
@ggreif Thanks !

rhine/src/FRP/Rhine/Clock/Realtime/Millisecond.hs
```diff
@@ -53,6 +53,17 @@ instance GetClockProxy (Millisecond n)
 --   and waits for the remaining time until the next tick.
 --   If the next tick should already have occurred,
 --   the tag is set to 'False', representing a failed real time attempt.
+--   Note that this clock internally uses 'threadDelay' which can block
```
turion, line 56:
```suggestion

--   Note that this clock internally uses 'threadDelay' which can block
```

rhine/src/FRP/Rhine/Clock/Realtime/Millisecond.hs
```diff
@@ -53,6 +53,17 @@ instance GetClockProxy (Millisecond n)
...
+--   for quite a lot longer than the requested time, which can cause
+--   the clock to miss one or more ticks when using low values of 'n'. 
+--   When using threadDelay, the difference between the real wait time 
+--   and the requested wait time will will be larger when using 
+--   the '-threaded' ghc option (around 800 microseconds) than when not using
+--   this option (around 100 microseconds). For low values of n it is recomended
```
turion, line 62:
```suggestion
--   this option (around 100 microseconds). For low values of n it is recommended
```

rhine/src/FRP/Rhine/Clock/Realtime/Millisecond.hs
```diff
@@ -53,6 +53,17 @@ instance GetClockProxy (Millisecond n)
 --   and waits for the remaining time until the next tick.
 --   If the next tick should already have occurred,
 --   the tag is set to 'False', representing a failed real time attempt.
+--   Note that this clock internally uses 'threadDelay' which can block
+--   for quite a lot longer than the requested time, which can cause
+--   the clock to miss one or more ticks when using low values of 'n'. 
+--   When using threadDelay, the difference between the real wait time
```
turion, line 59:
```suggestion
--   When using 'threadDelay', the difference between the real wait time 
```

rhine/src/FRP/Rhine/Clock/Realtime/Millisecond.hs
```diff
@@ -53,6 +53,17 @@ instance GetClockProxy (Millisecond n)
 --   and waits for the remaining time until the next tick.
 --   If the next tick should already have occurred,
 --   the tag is set to 'False', representing a failed real time attempt.
+--   Note that this clock internally uses 'threadDelay' which can block
+--   for quite a lot longer than the requested time, which can cause
+--   the clock to miss one or more ticks when using low values of 'n'. 
+--   When using threadDelay, the difference between the real wait time 
+--   and the requested wait time will will be larger when using
```
turion, line 60:
```suggestion
--   and the requested wait time will be larger when using 
```

rhine/src/FRP/Rhine/Clock/Realtime/Millisecond.hs
```diff
@@ -53,6 +53,17 @@ instance GetClockProxy (Millisecond n)
...
 --   the tag is set to 'False', representing a failed real time attempt.
+--   Note that this clock internally uses 'threadDelay' which can block
+--   for quite a lot longer than the requested time, which can cause
+--   the clock to miss one or more ticks when using low values of 'n'. 
+--   When using threadDelay, the difference between the real wait time 
+--   and the requested wait time will will be larger when using 
+--   the '-threaded' ghc option (around 800 microseconds) than when not using
+--   this option (around 100 microseconds). For low values of n it is recomended
+--   that '-threaded' not be used in order to miss less ticks. When a tick is
+--   missed the clock will attempt to tick as fast as possible in order to catch
```
turion, line 64: Maybe not quite as fast as possible, but rather as fast as necessary, meaning that if it missed by `m` milliseconds it will wait `m` milliseconds less.

turion (changes requested):
Looking great, thanks a lot! A few small improvements.

rhine/src/FRP/Rhine/Clock/Realtime/Millisecond.hs
miguel-negrao: What I mean here is that if  if a tick was missed because `threadDelay` waited too much (or the cell took too long to complete), then `remaining` becomes negative, which will cause `threadDelay` not to wait, executing the cell immediately. Is this accurate ? I'm assuming that the tick "happens" exactly when threadDelay returns (and the value of `now` is obtained).

miguel-negrao: Actually, I think `remaining` only becomes negative if the clock misses two ticks in a row. @turion , perhaps this is more clear ? "The clock will adjust the wait time, up to no wait time at all, to catch up when one or more ticks are missed."

turion: > Actually, I think remaining only becomes negative if the clock misses two ticks in a row.

I don't see that. 

> perhaps this is more clear 

That's a good formulation. I think "...when a tick is missed" will suffice as well.

miguel-negrao: > > Actually, I think remaining only becomes negative if the clock misses two ticks in a row.
> 
> I don't see that.

A diagram to show my reasoning:

imagem

miguel-negrao: If threadDelay sleeps (blue segment) during two ticks, then remaining will be negative I believe. If it sleeps during just one tick, remaining can still be negative if the computation time is long enough (green segment).

miguel-negrao:
I think I have now addressed all the issues.

rhine/src/FRP/Rhine/Clock/Realtime/Millisecond.hs
turion: Oh, I get it as well now, thanks :) I misunderstood "the clock misses two ticks in a row" for "the clock misses a tick two times".

Issue #162, opened 2022-05-05
Flake and GHC 9.2.2


jmatsushita:
Hi there,

I've been using nix flakes to setup my local development environment with cabal only and have used that setup to get rhine to work with GHC 9.2.2. Is that something you're interested in a PR for? I've been using this setup to work on the `terminal` and `vty` examples I mentioned in https://github.com/turion/rhine/issues/161

I'm also happy to try and break this down into 2 separate PRs (one to upgrade GHC, and another to port the project to flakes). It'll take a bit of work to make sure the CI also works properly as well as your local dev environment (I'm on an M1 mac).

Cheers,

Jun

turion:
Yes that sounds like a great idea! I just recently started using flakes, that's why I hadn't added anything like it yet.

> I'm also happy to try and break this down into 2 separate PRs

Yes, that's the best way to go. Thanks!

jmatsushita:
I gave a shot at a simple flake. Since I've worked on the 9.2.2 upgrade using flakes, I guess it makes sense to let the dust settle on that first :)

PR #164, opened 2022-05-21
Flake

jmatsushita:
Hi @turion 

Here's a very simple flake (following up on #162) which uses the experimental `haskell-flake-utils` [approach](https://github.com/ivanovs-4/haskell-flake-utils) which is the only tool I know that simulates someone a simple multi package setup with cabal (it doesn't do anything with the `cabal.project` file, instead it takes an argument where we can specify which packages needs to be built).

It allows to do 
```
nix build .#rhine-gloss
result/bin/rhine-gloss-gears
```

```
nix develop
cabal build rhine
#or 
cabal run rhine-gloss-gears
```

I've tested this on `aarch64` and `x86_64-linux` and the added CI job should test building on linux. Though I expect that `nix flake check` will fail with:
```
error: overlay does not take an argument named 'final'
```

(I've reported this bug to `haskell-flake-utils` https://github.com/ivanovs-4/haskell-flake-utils/issues/2

It builds using GHC 9.0.2 which is the default compiler in nixpkgs stable at the moment. It's not obvious how to build for [multiple GHC versions with this utility](https://github.com/ivanovs-4/haskell-flake-utils/issues/1) 

Some things to consider moving forward:
  - [x] Settle on `haskell-flake-utils` or a more manual approach.
  - [x] Update default.nix and shell.nix with `flake-compat` for legacy compatibility (as [described here](https://nixos.wiki/wiki/Flakes#Using_flakes_project_from_a_legacy_Nix))
  - [x] Remove other `*/shell.nix` and `*/default.nix` files. 
  - [x] Decide if maintaining stack build makes sense (I couldn't build with the current stack configuration)
  - [x] Update README
  - [x] ~~Add local development tooling (HLS, ghcid, hoogle...)~~ for later
  - [x] ~~Add a gitpod configuration to be able to have a cloud based development setup~~ for later
  - [x] ~~Find a solution for multi compiler builds~~ for later

Happy to keep adding to this PR and discuss as we go or merging smaller blocks of work, as you prefer.


flake.nix
```diff
@@ -0,0 +1,24 @@
...
+    nixpkgs.url = "github:NixOS/nixpkgs";
+    flake-utils.url = "github:numtide/flake-utils";
+    haskell-flake-utils.url = "github:ivanovs-4/haskell-flake-utils";
+    haskell-flake-utils.inputs.nixpkgs.follows = "nixpkgs";
+    haskell-flake-utils.inputs.flake-utils.follows = "flake-utils";
+
+  };
+
+outputs = { self, nixpkgs, flake-utils, haskell-flake-utils, ... }@inputs:
+  flake-utils.lib.eachSystem [ "x86_64-linux" "aarch64-darwin" ] (system:
```
turion, line 15: Why exactly these systems?

flake.nix
```diff
@@ -0,0 +1,24 @@
...
+    inputs.haskell-flake-utils.lib.simpleCabalProject2flake {
+      inherit self nixpkgs system;
+
+      name = "rhine";
+      packageNames = ["rhine-gloss" "rhine-examples"];
```
turion, line 20: Why not `"rhine"` as well? Also, could we get those names by reading the file system? Using `builtins.readDir` or similar?

flake.nix
```diff
@@ -0,0 +1,24 @@
...
+    flake-utils.url = "github:numtide/flake-utils";
+    haskell-flake-utils.url = "github:ivanovs-4/haskell-flake-utils";
+    haskell-flake-utils.inputs.nixpkgs.follows = "nixpkgs";
+    haskell-flake-utils.inputs.flake-utils.follows = "flake-utils";
+
+  };
+
+outputs = { self, nixpkgs, flake-utils, haskell-flake-utils, ... }@inputs:
+  flake-utils.lib.eachSystem [ "x86_64-linux" "aarch64-darwin" ] (system:
+    inputs.haskell-flake-utils.lib.simpleCabalProject2flake {
```
turion, line 16:
```suggestion
    haskell-flake-utils.lib.simpleCabalProject2flake {
```

flake.nix
turion, line 20:
```suggestion
      packageNames = [ "rhine-gloss" "rhine-examples" ];
```

flake.nix
```diff
@@ -0,0 +1,24 @@
...
+      inherit self nixpkgs system;
+
+      name = "rhine";
+      packageNames = ["rhine-gloss" "rhine-examples"];
+
```
turion, line 21:
```suggestion
```

turion:
https://github.com/turion/rhine/runs/6568785125?check_suite_focus=true fails with `error: overlay does not take an argument named 'final'`. The nix version used is 2.8.1.

turion:
> Here's a very simple flake (following up on #162) which uses the experimental `haskell-flake-utils` [approach](https://github.com/ivanovs-4/haskell-flake-utils) which is the only tool I know that simulates someone a simple multi package setup with cabal (it doesn't do anything with the `cabal.project` file, instead it takes an argument where we can specify which packages needs to be built).
> 
> It allows to do
> 
> ```
> nix build .#rhine-gloss
> result/bin/rhine-gloss-gears
> ```
> 
> ```
> nix develop
> cabal build rhine
> #or 
> cabal run rhine-gloss-gears
> ```

Awesome!

> I've tested this on `aarch64` and `x86_64-linux` and the added CI job should test building on linux. Though I expect that `nix flake check` will fail with:
> 
> ```
> error: overlay does not take an argument named 'final'
> ```

Indeed it does.

> (I've reported this bug to `haskell-flake-utils` [ivanovs-4/haskell-flake-utils#2](https://github.com/ivanovs-4/haskell-flake-utils/issues/2)

And it seems it's fixed upstream yesterday :tada: can you bump `haskell-flake-utils` and we try again?

> It builds using GHC 9.0.2 which is the default compiler in nixpkgs stable at the moment. It's not obvious how to build for [multiple GHC versions with this utility](https://github.com/ivanovs-4/haskell-flake-utils/issues/1)

That's good enough for a start :)

>     * [ ]  Settle on `haskell-flake-utils` or a more manual approach.

Let's see whether the upstream fix works. If yes, keep it, if no, a more manual approach.

>     * [ ]  Update default.nix and shell.nix with `flake-compat` for legacy compatibility (as [described here](https://nixos.wiki/wiki/Flakes#Using_flakes_project_from_a_legacy_Nix))

Yes makes sense :)

>     * [ ]  Remove other `*/shell.nix` and `*/default.nix` files.

I guess not, it contradicts the previous idea.

>     * [ ]  Decide if maintaining stack build makes sense (I couldn't build with the current stack configuration)

I don't care about stack at all, but maybe @miguel-negrao is motivated to fix it :grinning: 

>     * [ ]  Update README

Yes.

>     * [ ]  Add local development tooling (HLS, ghcid, hoogle...)

I don't know how HLS from the flake interacts with the editor. I'd leave these points out for now.

>     * [ ]  Add a gitpod configuration to be able to have a cloud based development setup

No idea what that is :sweat_smile: 

>     * [ ]  Find a solution for multi compiler builds

Yes, that can be a separate issue.

> Happy to keep adding to this PR and discuss as we go or merging smaller blocks of work, as you prefer.

I prefer smaller blocks. Too big PRs tend to get stuck.

flake.nix
jmatsushita: Another approach is to [use `eachDefaultSystem`](https://github.com/numtide/flake-utils/blob/master/default.nix#L3-L9) which is more comprehensive, but I went the route of picking the 2 architectures that I was able to test with.

flake.nix
jmatsushita: Yes good question. I thought `"rhine"` was implied since `packageNames` is optional. But I'll double check.

jmatsushita:
> Can you comment in the readme how to use the flake?

Sure thing will do.

Saw the upstream fix, and will bump it!

> Let's see whether the upstream fix works. If yes, keep it, if no, a more manual approach.

There are other concerns, but I'd like to merge this and submit another PR for discussing what other approaches would entail.

> > * [ ]  Remove other `*/shell.nix` and `*/default.nix` files.

> I guess not, it contradicts the previous idea.

I meant the `rhine/shell.nix` `rhine-examples/shell.nix` and so on. Unless there's something I'm missing in your workflow, it should be possible to enter the `nix develop` shell and build all packages together or individually.

> I don't know how HLS from the flake interacts with the editor. I'd leave these points out for now.

That's one of the reasons I use flakes and nix :D It works very well, but I agree let's do this in a subsequent PR :)

> > * [ ]  Find a solution for multi compiler builds

> Yes, that can be a separate issue.

Should work with this [other upstream fix](https://github.com/ivanovs-4/haskell-flake-utils/issues/1)

> I prefer smaller blocks. Too big PRs tend to get stuck.

Sounds good! 

Thanks a ton for the review, and super excited about your (conceptual) fixes on the `terminal` PR. Will fix this PR and look at the other one sometime this week 👍

turion:
> > Let's see whether the upstream fix works. If yes, keep it, if no, a more manual approach.
> 
> There are other concerns, but I'd like to merge this and submit another PR for discussing what other approaches would entail.

Great.

> > > * [ ]   Remove other `*/shell.nix` and `*/default.nix` files.
> 
> > I guess not, it contradicts the previous idea.
> 
> I meant the `rhine/shell.nix` `rhine-examples/shell.nix` and so on. Unless there's something I'm missing in your workflow, it should be possible to enter the `nix develop` shell and build all packages together or individually.

Oh, I see! Yes, let's remove those.

> > I don't know how HLS from the flake interacts with the editor. I'd leave these points out for now.
> 
> That's one of the reasons I use flakes and nix :D It works very well, but I agree let's do this in a subsequent PR :)

Ok, looking forward to that :)

flake.nix
turion: I'd propose either to use `eachDefaultSystem` or all systems we can somehow check on CI.

jmatsushita: > I'd propose either to use eachDefaultSystem or all systems we can somehow check on CI.

If that's OK, I would ask that we relax that requirement given that `aarch64-darwin` (M1 chip) is not available on Github Runners, but that's the machine architecture I'm using. 

What I can do is setup a CI matrix for systems that are available in Github that will build `x86_64-linux` and `x86_64-darwin`  (maybe try `i686-linux` too).

Note that with [this PR](https://github.com/ivanovs-4/haskell-flake-utils/pull/4) `haskell-flake-utils` changed their API such that now we pass an array of systems instead of using `flake-utils.lib.eachSystem` (which is now called in `haskell-flake-utils`).

flake.nix
jmatsushita: Yes indeed `"rhine"` [is added to the list of packages](https://github.com/ivanovs-4/haskell-flake-utils/blob/master/simpleCabalProject2flake.nix#L75):
```
 packageNames_ = pkgs: pkgs.lib.lists.unique ([ name ] ++ packageNames);
```

flake.nix
jmatsushita: Fixed

flake.nix
jmatsushita: Fixed

turion:
Awesome :) two systems built is already great!

Let's leave out `nix flake check` for the moment. The tests (when they're added) can be run on the cabal jobs for the time being.

flake.nix
```diff
@@ -0,0 +1,32 @@
...
+    flake-compat = {
+      url = "github:edolstra/flake-compat";
+      flake = false;
+    };
+
+    haskell-flake-utils.url = "github:ivanovs-4/haskell-flake-utils";
+    haskell-flake-utils.inputs.nixpkgs.follows = "nixpkgs";
+    haskell-flake-utils.inputs.flake-utils.follows = "flake-utils";
+
```
turion, line 17:
```suggestion
```

flake.nix
```diff
@@ -0,0 +1,32 @@
...
+  };
+
+outputs = { self, nixpkgs, flake-utils, haskell-flake-utils, flake-compat, ... }:
+  haskell-flake-utils.lib.simpleCabalProject2flake {
+    inherit self nixpkgs;
+
+    systems = [ "x86_64-linux" "x86_64-darwin"   # Tested in CI
+                "aarch64-darwin"                 # Tested by maintainers
+                "i686-linux"                     # Not sure we can test this in CI
+              ];
```
turion, line 27:
```suggestion
    systems = [
      "x86_64-linux" "x86_64-darwin"   # Tested in CI
      "aarch64-darwin"                 # Tested by maintainers
      "i686-linux"                     # Not sure we can test this in CI
    ];
```

shell.nix
```diff
@@ -1,4 +1,9 @@
-let
-  mypkgs = import ./. {};
-  shellFrom = import ./nix/shellFrom.nix {};
-in shellFrom mypkgs.rhine-gloss
+(import (
+  let
+    lock = builtins.fromJSON (builtins.readFile ./flake.lock);
+  in fetchTarball {
+    url = "https://github.com/edolstra/flake-compat/archive/${lock.nodes.flake-compat.locked.rev}.tar.gz";
+    sha256 = lock.nodes.flake-compat.locked.narHash; }
```
turion, line 6:
```suggestion
    sha256 = lock.nodes.flake-compat.locked.narHash;
  }
```

shell.nix
```diff
@@ -1,4 +1,9 @@
...
+) {
+  src =  ./.;
+}).shellNix
```
turion, line 9:
```suggestion
let
  lock = builtins.fromJSON (builtins.readFile ./flake.lock);
  flake-compat-src = fetchTarball {
    url = "https://github.com/edolstra/flake-compat/archive/${lock.nodes.flake-compat.locked.rev}.tar.gz";
    sha256 = lock.nodes.flake-compat.locked.narHash;
  };
  flake-compat = import flake-compat-src { src =  ./.; };
in
flake-compat.shellNix
```

I find so many big bracket pairs hard to parse

default.nix
```diff
@@ -1,12 +1,9 @@
...
-    rhine
-    rhine-gloss
-    rhine-examples
-  ;
-}
+(import (
+  let
+    lock = builtins.fromJSON (builtins.readFile ./flake.lock);
+  in fetchTarball {
+    url = "https://github.com/edolstra/flake-compat/archive/${lock.nodes.flake-compat.locked.rev}.tar.gz";
```
turion, line 5: Can you refactor this file like `shell.nix`?

default.nix
```diff
@@ -1,12 +1,9 @@
...
-  localPackages = import ./nix/localPackages.nix attrs;
-in
-{
-  inherit (localPackages)
-    rhine
-    rhine-gloss
-    rhine-examples
-  ;
-}
+(import (
```
turion, line 1: I think there is a way to detect in a `default.nix` whether it is called from `nix build` or `nix shell`. This could be used to remove `shell.nix` completely. But I can't find the option.

turion (changes requested):
Looking great :) just some small syntax things.

default.nix
jmatsushita: If we're ok with requiring a recent version of nix (from `22.05` it's the default, and before it requires setting `experimental-features = nix-command flakes`, then we actually need neither `default.nix` nor `shell.nix` (which are only here to [support legacy nix](https://nixos.wiki/wiki/Flakes#Using_flakes_project_from_a_legacy_Nix)).

turion: Yes, that's also ok. Let's drop support for legacy nix.

flake.nix
```diff
@@ -0,0 +1,36 @@
...
+    flake-utils.url = "github:numtide/flake-utils";
+
+    flake-compat = {
+      url = "github:edolstra/flake-compat";
+      flake = false;
+    };
+
+    haskell-flake-utils.url = "github:ivanovs-4/haskell-flake-utils";
+    haskell-flake-utils.inputs.nixpkgs.follows = "nixpkgs";
+    haskell-flake-utils.inputs.flake-utils.follows = "flake-utils";
```
turion, line 16:
```suggestion
    haskell-flake-utils = {
      url = "github:ivanovs-4/haskell-flake-utils";
      inputs.nixpkgs.follows = "nixpkgs";
      inputs.flake-utils.follows = "flake-utils";
    };
```

default.nix
turion: Or rather, I'll leave the choice up to you. Whatever you're more comfortable with.

jmatsushita:
That's great. Will open a new PR for adding haskell language server and look into the multi compiler setup. 👍

Issue #161, opened 2022-05-02
Rhine and Terminal


jmatsushita:
Hi @turion, 

I'm finally spending some time poking at Rhine which I wanted to do for a while. I really love the concept and I'm bumping into some questions and issues that you probably will be able to help me with!

I'm trying to integrate the `haskell-terminal`  library with Rhine, trying to replace the `StdinClock` with a clock that returns terminal events (and a subclock that filters on key events). As you'll see I'm also trying to compose a periodic clock with an IO event driven clock, as well as printing to screen. I have tried to think about the display/actuator side as a Clock, but I couldn't wrap my head around how to do this (maybe by passing state to the clock and flushing to the terminal periodically?). Anyway that's not really the problem I wanted to ask help for, but I thought I'd also mention this in passing.

The issue I'm running into when running this toy program https://gist.github.com/jmatsushita/d94f04479666979b57d99558fcf3f40d/49ac895e92fa620953086cd734cff21af371df68 is that I get the message 
```
thread blocked indefinitely in an STM transaction
```

I'm a bit confused about how I should go about hoisting the TerminalT monad transformer into IO. I probably have the wrong idea running `withTerminal $ runTerminalT` twice, in `initClock` (which installs the `awaitEvent` handler) and in `liftTerminal` which tries to hoist the ClSF on the display side, but I've wrestled with it for a while now and I can't wrap my head around how to make this work.

Hoping this could be useful as an example of how to use clocks that have effects in transformers, and I realise that hoisting `Rhine`s is a more difficult work in progress, but it doesn't seem like I should need that, but I might be wrong.

Anyway, thanks a ton for the awesome paper and library, really looking forward to get unstuck :)

Jun

jmatsushita:
I should probably study this https://github.com/turion/rhine/blob/master/rhine-gloss/src/FRP/Rhine/Gloss/IO.hs :)

turion:
Thanks for your kind words :) this is a great idea!

I still don't understand the `terminal` library well enough to understand the error exactly. What I don't understand at all: `withTerminal :: (MonadIO m, MonadMask m) => (System.Terminal.Platform.LocalTerminal -> m a) -> m a` expects a function, but you give it the output of `runTerminalT`. So you never use the value of type `System.Terminal.Platform.LocalTerminal`. So how do you ensure that the clock and the `ClSF` are on the same terminal?

[...]

PR #165, opened 2022-05-21
rhine-terminal backend with haskell-terminal library and Repl like example

jmatsushita:
As per https://github.com/turion/rhine/issues/161 this is my attempt at using the `haskell-terminal` library with `rhine` and succeeding very moderately, as this works dreadfully slowly.  

I've stacked the PR on top of my `flake` branch so it's easy to build and run this way:

```
nix build .#rhine-terminal
result/bin/rhine-terminal-repl
```

Running this should show a prompt, with a blinking `.` before a `>` and a terminal cursor after it. When pressing keys, they appear but after a long delay. This slow behavior exists both on `aarch64` and `x86`. Let me know if you can reproduce :) 

I haven't yet split out the Terminal related functionality into a module like `FRP.Rhine.Gloss` yet as we work through the performance problems!


turion:
https://github.com/turion/rhine/runs/6568785125?check_suite_focus=true fails with `error: overlay does not take an argument named 'final'`. The nix version used is 2.8.1.

turion:
Merge #164 first.

rhine-terminal/Main.hs
```diff
@@ -0,0 +1,135 @@
...
+import Control.Monad.Schedule ()
+
+import Data.Text (Text)
+import qualified Data.Text as T
+import System.Terminal
+import System.Terminal.Internal
+import FRP.Rhine
+import Control.Monad.Trans.Class (lift)
+import Control.Monad.Trans.Reader (ReaderT)
+import Data.List (singleton)
```
turion, line 33: This import seems to be unused.

turion:
I've given this a spin. No idea what it's supposed to do exactly.

* It just worked with `nix develop`. awesome!
* CPU usage is at 1% or less, which is good.
* It seems that if I type several characters, they're buffered and only one is printed at a time. Is that what you mean by it being slow? I'll have a look, maybe I can figure it out.
* There's no way to exit the repl. `ctrl+d` or `esc` would be great.

turion:
The issue you had was that your `display` component was on the `MilliSecond 1000` clock. That means that display is only called every second. Which means it only collects one item do display every second. I changed it slightly so that the prompt beat is on the 1000 ms clock, and the key display is on the terminal clock. Now everything works like a charm it seems.

turion:
Ironically, this is exactly the kind of bug `rhine` is supposed to resolve :laughing:

turion (changes requested):
Alright, so this seems in good shape for me to go forward! It seems `terminal` can be used as a rhine backend. Some todos to proceed:

* [x] Add `rhine-terminal` to `cabal.project` and `stack.yaml`.
* [x] Factor out the terminal clock to a library.
* [x] Keep the executable as an example application.
* [ ] If you're motivated add an integration test that produces some characters on stdout.

jmatsushita:
> Ironically, this is exactly the kind of bug rhine is supposed to resolve 😆 (but at least it was clear to see and reason about)

Ha ha ha, well I suspected the bug was in my mis-use of rhine :) Looking forward to dig into this. Super grateful for being unblocked 🙏

jmatsushita:
Hi @turion ,

So I got started on the todo list:

> * [x]  Add `rhine-terminal` to `cabal.project` and `stack.yaml`.

I've added `rhine-terminal` to the `flake.nix`. If you want we can change the `cabal.project` file to list the packages explicitly. For `stack.yaml` it might make more sense to have https://github.com/turion/rhine/pull/166 merged beforehand?

> * [x]  Factor out the terminal clock to a library.
> * [x]  Keep the executable as an example application.

I've made good progress I think and followed the `rhine-gloss` approach to try and avoid threading `term` throughout using a ReaderT, but I got myself into another corner and the app is not running anymore :/ When starting it, nothing happens... Would appreciate your eyes on this.

> * [ ]  If you're motivated add an integration test that produces some characters on stdout.

Happy to try when the example works!

### Some questions about ways to compose rhines and signal networks

With regards to your fix and my misunderstanding, I see that you refactored the Rhines such that they are composed with "Time-parallel signal composition", using `||@ terminalConcurrently @||`. I've tried to represent this here:

image

However this leaves me wondering whether it's possible to fix the problem while keeping the original intent composing sensors and actuators like so:

image

Should there be a way to compose Rhines :

```haskell
compRhines :: Rhine m cl a b -> Rhine m cl b c -> Rhine m cl a c
```

I understand from the paper that Rhines and Signal Networks don't form categories in general, but it seems that when clocks have the same values (or maybe the same tree structure?) it should be possible to compose Rhine end to end like that?

Another way to ask the same question I suppose is: is there currently a way to consume a `Rhine m (ParallelClock m clL clR) a (Either b c)` ?

I've committed the application code with the refactor and added the `compRhines` shenannigans at the bottom, hoping there's a solution down that path too.

.envrc
```diff
@@ -0,0 +1,2 @@
+[[ -f .envrc.local ]] && source_env .envrc.local
+use flake
```
turion, line 2: What does this do?

rhine-terminal/src/FRP/Rhine/Terminal.hs
```diff
@@ -0,0 +1,165 @@
...
+import Data.IORef ()
+
+-- time
+import Data.Time.Clock ( getCurrentTime )
+
+-- transformers
+import Control.Monad.Trans.Class ( MonadTrans )
+import Control.Monad.Trans.Reader ( ReaderT(..), ask )
+
+-- terminal
```
turion, line 32:
```suggestion
```

rhine-terminal/src/FRP/Rhine/Terminal.hs
```diff
@@ -0,0 +1,165 @@
...
+  deriving (Functor, Applicative, Monad, MonadTrans, MonadIO, MonadThrow, MonadCatch, MonadMask)
+
+instance MonadInput (TerminalConcT IO) where
+  awaitWith = awaitWith
+
+instance MonadPrinter (TerminalConcT IO) where
+  putLn = putLn
+  putChar = putChar
+  putText = putText
+  getLineWidth = getLineWidth
```
turion, line 67:
```suggestion
  awaitWith = awaitWith

instance MonadPrinter (TerminalConcT IO) where
  putLn = putLn
  putChar = putChar
  putText = putText
  getLineWidth = getLineWidth
```

I don't get these instances. They look like infinite loops to me. Don't you want to wrap/unwrap the newtype? Also you could try to derive `MonadPrinter` etc., does that not work with `GeneralizedNewtypeDeriving`?

rhine-terminal/src/FRP/Rhine/Terminal.hs
```diff
@@ -0,0 +1,165 @@
...
+instance Semigroup TerminalEventClock where
+  _ <> _ = TerminalEventClock
+
+-- * Reactimation
+
+launchTerminal
+  :: MonadIO    m
+  => LocalTerminal
+  -> TerminalConcT m a
+  ->            m a
```
turion, line 124:
```suggestion
  :: MonadIO       m
  => LocalTerminal
  -> TerminalConcT m a
  ->               m a
```
I guess this was your intention?

rhine-terminal/src/FRP/Rhine/Terminal.hs
```diff
@@ -0,0 +1,165 @@
...
+  hideCursor = hideCursor
+  setAutoWrap = setAutoWrap
+  setAlternateScreenBuffer = setAlternateScreenBuffer
+
+-- * Terminal clock in 'IO'
+
+-- | Concurrently block on @gloss@ events.
+data TerminalEventClock = TerminalEventClock
+
+instance (MonadIO m, MonadMask m) => Clock (TerminalConcT m) TerminalEventClock where
```
turion, line 98: Can you leave the `TerminalConcT` out of the clock definition? 
```suggestion
instance (MonadIO m, MonadMask m, MonadReader Terminal m, MonadInput m) => Clock m TerminalEventClock where
```

This has the advantage that we don't need to pin down the transformer stack when using the clock.

rhine-terminal/src/FRP/Rhine/Terminal.hs
turion: Ah we can't derive directly because `ReaderT` has no `MonadInput` instances and so on. So you need to write them by hand unfortunately.

rhine-terminal/src/FRP/Rhine/Terminal.hs
```diff
@@ -0,0 +1,165 @@
...
+      UTCTime )
+import Control.Monad.Catch (MonadMask, MonadCatch, MonadThrow)
+
+-- | Wraps the LocalTerminal value needed for communication with the @terminal@ backend.
+newtype TerminalConcT m a = TerminalConcT
+  { unTerminalConcT :: ReaderT LocalTerminal m a }
+  deriving (Functor, Applicative, Monad, MonadTrans, MonadIO, MonadThrow, MonadCatch, MonadMask)
+
+instance MonadInput (TerminalConcT IO) where
+  awaitWith = awaitWith
```
turion, line 61:
```suggestion
  awaitWith callback = TerminalConcT $ ReaderT $ const $ awaitWith callBack
```
Or similar.

rhine-terminal/Main.hs
```diff
@@ -0,0 +1,137 @@
...
+
+-- Rhine of sinks
+actuate' :: Rhine App DisplayClock (Either Text Char) ()
+actuate' = display' @@ displayClock
+
+-- Should it be possible to compose rhines with the same clock values?
+-- We don't have clock equality though...
+
+compRhines :: Monad m => Rhine m cl a b -> Rhine m cl b c -> Rhine m cl a c
+compRhines (Rhine sn1 cl) (Rhine sn2 _) = case (sn1, sn2) of
```
turion, line 131: Here you're loosing the second clock. You should at least use `Monoid cl` for this to be meaningful.

rhine-terminal/Main.hs
```diff
@@ -0,0 +1,137 @@
...
+  (Synchronous clsf1, Synchronous clsf2) -> Rhine (Synchronous (clsf1 >-> clsf2)) cl
+  _ -> undefined -- dont' understand clock trees enough yet
```
turion, line 133: Add constraints `cl ~ In cl ~ Out cl` so the other cases will never occur.

rhine-terminal/Main.hs
```diff
@@ -0,0 +1,137 @@
...
+
+
+mainRhine' :: Rhine App DisplayClock () ()
+mainRhine' = sensor' `compRhines` actuate'
```
turion, line 137: This should crash because `sensor'` is not `Synchronous`.

turion:
> > * [x]   Add `rhine-terminal` to `cabal.project` and `stack.yaml`.
> 
> I've added `rhine-terminal` to the `flake.nix`.

:+1:

> If you want we can change the `cabal.project` file to list the packages explicitly.

No, the fewer places where we need to list things, the better.

> For `stack.yaml` it might make more sense to have #166 merged beforehand?

Yes, but that's not blocking.

> > * [x]   Factor out the terminal clock to a library.
> > * [x]   Keep the executable as an example application.
> 
> I've made good progress I think and followed the `rhine-gloss` approach to try and avoid threading `term` throughout using a ReaderT, but I got myself into another corner and the app is not running anymore :/ When starting it, nothing happens... Would appreciate your eyes on this.

I believe it's your `MonadInput` etc. instances which are basically infinite loops as far as I understand.

> > * [ ]   If you're motivated add an integration test that produces some characters on stdout.
> 
> Happy to try when the example works!

:+1: 

> ### Some questions about ways to compose rhines and signal networks
> 
> With regards to your fix and my misunderstanding, I see that you refactored the Rhines such that they are composed with "Time-parallel signal composition", using `||@ terminalConcurrently @||`. I've tried to represent this here:
> 
> image
> 
> However this leaves me wondering whether it's possible to fix the problem while keeping the original intent composing sensors and actuators like so:
> 
> image

Awesome graphics :) yes, this is also how I'd look at them I think.

> Should there be a way to compose Rhines :
> 
> ```haskell
> compRhines :: Rhine m cl a b -> Rhine m cl b c -> Rhine m cl a c
> ```

No, this doesn't work, for two reasons.

1. Clocks in rhine have identity. These are called "clock values". A signal network doesn't contain a clock (although it requires a specific clock type), but a `Rhine` does. So you'd force the two clocks in these `Rhine`s to be identical, although they can be different values of the same type. This would at most work with `Monoid cl`.
2. Even then `cl` might be composed of several clocks, which means that `b` leaves the first `Rhine` at a different rate than it enters the second one. This cannot work without a buffer. At most it would work if `cl ~ In cl ~ Out cl`, i.e. if `cl` is atomic.

If we make both assumptions, we can write something like your `compRhines`.

> Another way to ask the same question I suppose is: is there currently a way to consume a `Rhine m (ParallelClock m clL clR) a (Either b c)` ?

Huh, it seems you've hit a bullseye. It should morally be possible to precompose it with a [`BehaviourF m (Either b c) ()`](https://hackage.haskell.org/package/rhine-0.7.1/docs/FRP-Rhine-ClSF-Core.html#t:BehaviourF). I was so sure that there is a function that does that, but it seems there isn't. Definitely a missing feature. We should open a new ticket for that.

You can definitely postcompose with a pure function: https://hackage.haskell.org/package/rhine-0.7.1/docs/src/FRP.Rhine.SN.Combinators.html#%3E%3E%3E%5E It should be easy to generalise this to a Kleisli arrow `a -> m b`, in case you want to add a new simple combinator. But it's not so trivial to generalise to an `MSF`, I want to try that myself.

> I've committed the application code with the refactor and added the `compRhines` shenannigans at the bottom, hoping there's a solution down that path too.

Not in general I believe.

.envrc
jmatsushita: `direnv` allows to load the flake when `cd`ing into the directory. It's very convenient, but of course I shouldn't have smuggled it inside this PR :D Happy to remove it and put it into a follow up PR on the flakes PR (which I'll look at soon).

rhine-terminal/src/FRP/Rhine/Terminal.hs
jmatsushita: > They look like infinite loops to me. Don't you want to wrap/unwrap the newtype?

I thought I was doing something non-sensical here 😅

rhine-terminal/src/FRP/Rhine/Terminal.hs
jmatsushita: Yes, that makes a lot of sense.

jmatsushita: I am running into some trouble further down though when trying to write `terminalConcurrently` in this style because concurrently pins `IO`. 
```haskell
concurrently
  :: ( Clock IO cl1, Clock IO cl2
     , Time cl1 ~ Time cl2
     )
  => Schedule IO cl1 cl2
```

Should there be (can there be) a concurrently which is over `MonadIO m`?

rhine-terminal/Main.hs
turion, line 137: Can you instead rebase onto https://github.com/turion/rhine/pull/167 and use the new combinators from there to make something like `sensor' @>-^ display'`?

turion:
We cannot do MonadIO here. UnliftIO works in principle, but I don't think it's worth the dependency. I'm working on another solution to concurrency currently which removes the need for schedules, so we shouldn't put too much effort in that direction.
-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

rhine-terminal/src/FRP/Rhine/Terminal.hs
```diff
@@ -0,0 +1,165 @@
...
+  eraseInDisplay = eraseInDisplay
+  showCursor = showCursor
+  hideCursor = hideCursor
+  setAutoWrap = setAutoWrap
+  setAlternateScreenBuffer = setAlternateScreenBuffer
+
+-- * Terminal clock in 'IO'
+
+-- | Concurrently block on @gloss@ events.
+data TerminalEventClock = TerminalEventClock
```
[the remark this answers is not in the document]
turion: I'm thinking more and more that this is the right way. If you put the `LocalTerminal` in the clock, you could drop the `TerminalConcT` monad, make an `instance Clock IO TerminalEventClock`, and you don't need to write your own `Schedule` anymore. Yes, this would restrict the programs we can write now somewhat, but I think it's ok to start with until the dust has settled around other developments like https://github.com/turion/rhine/pull/148.

rhine-terminal/src/FRP/Rhine/Terminal.hs
jmatsushita: Ok sounds good. I had the feeling it wasn't worth the extra dependency.

rhine-terminal/Main.hs
```diff
@@ -0,0 +1,137 @@
...
+beat' :: ClSF App (LiftClock IO TerminalConcT BeatClock) () Text
+beat' = flip T.cons " > " . (cycle " ." !!) <$> count
+
+key' :: ClSF App KeyClock () Char
+key' = tagS
+
+type DisplayClock' = ParClock App (LiftClock IO TerminalConcT BeatClock) KeyClock
+
+displayClock :: DisplayClock
+displayClock = ParallelClock (liftClock waitClock) keyClock terminalConcurrently
```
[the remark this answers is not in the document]
jmatsushita: Yes happy to :)

rhine-terminal/src/FRP/Rhine/Terminal.hs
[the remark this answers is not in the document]
jmatsushita: Great! Happy to follow this path for now and follow new developments.

jmatsushita:
Ok I believe I made further progress, and the example does execute properly with the 2 approaches, but I'm sure I'm still not quite there yet:
- I've added a `LocalTerminal` to the `TerminalEventClock`, and it makes writing the instance easy, but I have to thread `term` through which is not great. (Is the fact that `TerminalT`, while a ReaderT under the hood, doesn't export its constructor causing trouble here?)
- I'll try and use the `Terminal t` constraint instead of using `LocalTerminal` as I think it would help testing (with [`VirtualTerminal`](https://hackage.haskell.org/package/terminal-0.2.0.0/docs/System-Terminal-Internal.html#g:3))
- instead of an `instance Clock IO TerminalEventClock` I have an `instance Clock (TerminalT LocalTerminal IO) TerminalEventClock` otherwise I struggle with being able to run `runTerminalT` once at the top and providing the `MonadScreen m` constraints. I get the feeling I'm missing something here. 
- the semigroup instance for `TerminalEventClock` seems wrong. But how can we write one?
- I seem to still need a `terminalConcurrently term` in the `mainRhine`. Should I try and use `concurrently` here instead? It seems that in that case I have to renounce calling `runTerminalT` at the top level?

The alternate approach with post-composing the ClSF works (ce10eca)! I'm wondering now if there's a way to make this construction more "symmetric". It seems arbitrary somehow that the `Rhine` is the `sensor` while the `display` is the `ClSF`, why not the other way around? I understand that the idea is that the Rhine has a clock value, so we make sure we can only compose   I suppose "symmetry" could be "restored" if:
    - I could instead build `sensor'` as a `ClSF`, compose it with `display` and then only later make a `Rhine`? I would need to be able to run `beat` and `key` in parallel as `ClSF`s but that requires a `SN`, so we're back to post-composing the `display` `ClSF` with the `SN` but now with `>--^` instead of `@>-^` which still feels "asymmetric" to me.
    - or is it possible to assemble `ClSF`s into a top level `SN` and then separately build a clock tree value that we then apply to build a `Rhine`? 

Looking forward to your always helpful and edifying comments!

jmatsushita:
I tried tests but I'm not sure what to do [here](https://github.com/turion/rhine/pull/165/commits/29b30472d7cbcd87f8a5c769a111ee11e45f51c1#diff-d13071753714ea29c00701c2ff40429dfe0afd17d7beabb6c2f404b7a8e011dfR51):
```
void $ liftIO $ forkIO $ runTerminalT (flow $ testRhine t) t
```

If I don't `forkIO` then I get the error
```
rhine-terminal-tests: thread blocked indefinitely in an MVar operation
```

I have the feeling I shouldn't use `flow` here...

jmatsushita:
I've added a few more features to the example e212594, following the "alternate" approach made possible by the Rhine ClSF post-composition and refactoring the app with that style. 

So good things are:
 - It now exits on Control-C
 - It supports enter and backspace (but not cursor movement yet)

But on the negative side:
 - It seems very sluggish and sometimes misses input
 - It sometimes displays `J` when hitting enter

Hope you can take a look and advise 🙏

turion:
I think the main problem is that you're using a `concurrently` schedule for both `SignalClock` and `InputClock`. This can't work: It starts two listeners on each clock, and the GHC runtime decides randomly which will be activated with the next event. Meaning that sometimes the `SignalClock` will swallow the input and discard it.

Instead you need to use `schedSelectClocks` to schedule two `SelectClock`s. It will make sure that the main clock will only be started once, and all the events broadcast to each subsystem.

I also get this funny `J` popping up each time I press return, no idea where that is coming from.

turion:
I took the liberty and added two commits to your branch fixing this.

turion:
> I tried tests but I'm not sure what to do [here](https://github.com/turion/rhine/pull/165/commits/29b30472d7cbcd87f8a5c769a111ee11e45f51c1#diff-d13071753714ea29c00701c2ff40429dfe0afd17d7beabb6c2f404b7a8e011dfR51):
> 
> ```
> void $ liftIO $ forkIO $ runTerminalT (flow $ testRhine t) t
> ```
> 
> If I don't `forkIO` then I get the error
> 
> ```
> rhine-terminal-tests: thread blocked indefinitely in an MVar operation
> ```
> 
> I have the feeling I shouldn't use `flow` here...

No, this is fine. `flow` is blocking, so the test wouldn't be productive if you have `flow` in the main thread.

jmatsushita:
> Instead you need to use schedSelectClocks to schedule two SelectClocks. 

Awesome, I understand now! Maybe `schedSelecClocks` should have `concurrently` in its name given this pattern:
```
  ( ( input ++@ schedSelectClocks @++ signal )
            ++@ concurrently @++ prompt
```

> I took the liberty and added two commits to your branch fixing this.

Thanks, glad you did!

> I'm still not convinced this is the best layout for the app.

That's interesting. I think I'm using this style because indeed as you pointed, I'm envisioning something where state is shared between components and that's not reflected in the current example. Maybe for educational purposes it makes sense to have a simpler example where things are wired as you suggested, and another which has shared state? I've been wondering whether examples could go in the `rhine-examples` package, a `TerminalSimple.hs` example  and a `TerminalStateful.hs` example could help to demonstrate different patterns?

> This is only necessary if the different components need to share some state somehow

Is there a structure here where 
```
beat >-> displayBeat ||@ concurrently @|| key >-> displayKey
```
can be "distributed" under some condition as 
```
beat ++@ concurrently @++ key @>-^ (displayBeat ?? displayKey)
```

Where `??` is the case multiplexing (an `||` for consuming multiplexed sources). I think that's the essence of my diagram from before, and even though I agree in my example there's no reason to lay it out like that, I'm trying to make a mental model of how I can compose sources and sinks in that way (and it's also useful to know when I shouldn't or cannot).

Also in your refactor, you "pushed down" the `runTerminalT` from the top level, to inside `display`. It makes me see a pattern, and I wonder if it makes sense to see things this way. we have a `runTerminalT` inside the `TerminalEventClock`, therefore in both `SelectClock` that are made from it, and "in" the 2 sources which depend on terminal effects. Then on the other side, in the `display` sink which also depends on terminal effects.

I'm wondering:
 - Is that why BeatClock lost its `LiftClock IO AppT`, as now all sources run in `IO` because we pushed `runTerminalT` down?
 - Does this mean `terminalConcurrently` is actually not needed anymore? Or should we keep it to allow arranging the monad transformer stack differently?
 - Could this be done in an extensible effect kind of way? Where we'd declare the effects of a part of the clock regions, and "consume" them elsewhere, to keep track of them? I suppose I'd like to see more clearly in the types that the effects of `beat` are different than the effects of `key`, but maybe that doesn't make sense :) 

> No, this is fine. flow is blocking, so the test wouldn't be productive if you have flow in the main thread.

Ok, sorry but I'm still not sure how to make the tests work, any other hints that could help make these tests pass?

jmatsushita:
Also, I'm happy that we try to converge towards a simple example in order to merge and discuss other topics in another PR or even outside this repo if you'd prefer. Just let me know :)

turion:
> > I'm still not convinced this is the best layout for the app.
> 
> That's interesting. I think I'm using this style because indeed as you pointed, I'm envisioning something where state is shared between components and that's not reflected in the current example.

Ah, yes then you're on the right track!

> Maybe for educational purposes it makes sense to have a simpler example where things are wired as you suggested, and another which has shared state? I've been wondering whether examples could go in the `rhine-examples` package, a `TerminalSimple.hs` example and a `TerminalStateful.hs` example could help to demonstrate different patterns?

That's a very good idea. `TerminalSimple.hs` could be the example application inside `rhine-terminal`, and a bigger example (potentially using also, say, the gloss backend or some other clocks) could reside in `rhine-examples`. It's also possible to put two example applications in `rhine-terminal`, if they only use the terminal.

> Also, I'm happy that we try to converge towards a simple example in order to merge and discuss other topics in another PR or even outside this repo if you'd prefer. Just let me know :)

Yes, that would be great. How about putting the simple example in this PR so we are sure that the library works, and a more elaborate one in a second PR?

Also, before that second PR I might merge #171 which will simplify the schedule story, which might also simplify the bigger second example application.

> > This is only necessary if the different components need to share some state somehow
> 
> Is there a structure here where
> 
> ```
> beat >-> displayBeat ||@ concurrently @|| key >-> displayKey
> ```
> 
> can be "distributed" under some condition as
> 
> ```
> beat ++@ concurrently @++ key @>-^ (displayBeat ?? displayKey)
> ```
> 
> Where `??` is the case multiplexing (an `||` for consuming multiplexed sources). I think that's the essence of my diagram from before, and even though I agree in my example there's no reason to lay it out like that, I'm trying to make a mental model of how I can compose sources and sinks in that way (and it's also useful to know when I shouldn't or cannot).

I think that should be [`(+++)`](https://hackage.haskell.org/package/base-4.16.1.0/docs/Control-Arrow.html#v:-43--43--43-) or `(|||)` from `ArrowChoice`. (`ClSF`s are `Arrow`s.) If you like arrow notation, you can also do something like:

```haskell
displayBeatOrKey = proc input do
  advanceJointState -< input
  case input of
    Left beat -> displayBeat -< beat
    Right key -> displayKey -< key
```

> Also in your refactor, you "pushed down" the `runTerminalT` from the top level, to inside `display`. It makes me see a pattern, and I wonder if it makes sense to see things this way. we have a `runTerminalT` inside the `TerminalEventClock`, therefore in both `SelectClock` that are made from it, and "in" the 2 sources which depend on terminal effects. Then on the other side, in the `display` sink which also depends on terminal effects.
> 
> I'm wondering:
> 
>     * Is that why BeatClock lost its `LiftClock IO AppT`, as now all sources run in `IO` because we pushed `runTerminalT` down?

Yes. This makes the type signatures simpler for now.

>     * Does this mean `terminalConcurrently` is actually not needed anymore? Or should we keep it to allow arranging the monad transformer stack differently?

Yes, it should not be needed. (And after #171 it will not be needed at all.)

>     * Could this be done in an extensible effect kind of way? Where we'd declare the effects of a part of the clock regions, and "consume" them elsewhere, to keep track of them? I suppose I'd like to see more clearly in the types that the effects of `beat` are different than the effects of `key`, but maybe that doesn't make sense :)

Yes, that's a good way to do this. Ideally the clock instances are all of the form

```haskell
instance MonadFoo m => Clock m FooClock where
  initClock = foo ...
```

Then we never need to specify `m` exactly while we compose `Rhine`s (thus avoiding `LiftClock` etc.). Then at the end after we do `eraseClock` or `flow` (and thus leave the clock types), we can handle the effects by instantiating with a particular monad or transformer, and running it.

An example for that approach might be https://github.com/turion/rhine/pull/171/files#diff-aff02ce436bc3b78434667d310d71ec7b7da917c52512daa8e055831346673e1, but there I use `eraseClock` instead of `flow`, which may be confusing.

> > No, this is fine. flow is blocking, so the test wouldn't be productive if you have flow in the main thread.
> 
> Ok, sorry but I'm still not sure how to make the tests work, any other hints that could help make these tests pass?

Ah sorry, I wasn't aware that they're failing. I'll look into it now.

turion:
I believe the reason the test is failing is because the virtual terminal doesn't emit events when executing the `PutText` command.

I think you'd need to do something else instead. Probably starting `flow rhineTerminal` in a thread and then maybe use `hPut` or so to simulate input? I haven't tried this, maybe it is necessary to instead start a process with hackage.haskell.org/package/process that produces input.

If testing is too complicated, let's put it in another PR. As you might have noticed, there are no tests for the other packages either, for similar reasons.

turion:
As a simple alternative, we can emulate the key event with a  `TQueue`. Example commit attached. I think this is the best option for now since it works and is simple.

jmatsushita:
Hi @turion 

> Yes, that would be great. How about putting the simple example in this PR so we are sure that the library works, and a more elaborate one in a second PR?

I kept the more involved example out of this PR as you suggested, and kept a version that's closer in spirit to the fixes you had initially made to my original code. I've submitted the other example in its current state as a draft PR and will continue building onto it, hopefully with your help.

> Also, before that second PR I might merge https://github.com/turion/rhine/pull/171 which will simplify the schedule story, which might also simplify the bigger second example application.

Such a great simplification! I looked at `monad-schedule` and feel like it's a really promising abstraction! Looking forward to give it a spin :)

> I think that should be (+++) or (|||) from ArrowChoice.

That's a good suggestion, I'll play with Arrow notation in the second example. 

> Ideally the clock instances are all of the form
> 
> instance MonadFoo m => Clock m FooClock where
>  initClock = foo ...

Yeah I'd like to revisit this, as it feels that we should be able to use the `MonadTerminal` constraint here.

> As a simple alternative, we can emulate the key event with a TQueue. Example commit attached. I think this is the best option for now since it works and is simple.

That's really great 👏  Thanks a ton.

jmatsushita:
I believe these are now all completed:
 - [x] Add rhine-terminal to cabal.project and stack.yaml.
 - [x] Factor out the terminal clock to a library.
 - [x] Keep the executable as an example application.
 - [x] If you're motivated add an integration test that produces some characters on stdout.
 
 I've made a few cosmetic changes too, so feel free to re-review in depth 👍

jmatsushita:
Oh I just realised there is a loose end still:
> the semigroup instance for TerminalEventClock seems wrong. But how can we write one?

```haskell
instance Terminal t => Semigroup (TerminalEventClock t) where
  t <> _ = t
```

What does it mean and can we do better?

cabal.project
```diff
@@ -1 +1,5 @@
-packages: */*.cabal
```
turion, line 1: Actually, the wildcard works well, doesn't it? I'd rather leave it if it doesn't make any trouble.

rhine-terminal/TerminalSimple.hs
```diff
@@ -0,0 +1,100 @@
...
+
+import Data.Text (Text)
+import qualified Data.Text as T
+import System.Terminal
+import System.Terminal.Internal
+import FRP.Rhine
+import System.IO hiding (putChar)
+
+import FRP.Rhine.Terminal (TerminalEventClock(..))
+import System.Exit (exitSuccess)
```
turion, line 26:
```suggestion
-- base
import Prelude hiding (putChar)
import System.Exit (exitSuccess)
import System.IO hiding (putChar)

-- text
import Data.Text (Text)
import qualified Data.Text as T

-- terminal
import System.Terminal
import System.Terminal.Internal

-- rhine
import FRP.Rhine

-- rhine-terminal
import FRP.Rhine.Terminal (TerminalEventClock(..))
```

I like imports grouped by packages.

rhine-terminal/TerminalSimple.hs
```diff
@@ -0,0 +1,100 @@
...
+
+-- Clocks
+
+data Input = Char Char
+           | Space | Backspace | Enter
+           | Exit
```
turion, line 32:
```suggestion
data Input
  = Char Char
  | Space
  | Backspace
  | Enter
  | Exit
```

rhine-terminal/TerminalSimple.hs
```diff
@@ -0,0 +1,100 @@
...
+
+type InputClock = SelectClock (TerminalEventClock LocalTerminal) Input
+
+inputClock :: LocalTerminal -> InputClock
+inputClock term = SelectClock { mainClock = TerminalEventClock term, select = selectKey }
+  where
+    selectKey :: Tag (TerminalEventClock LocalTerminal) -> Maybe Input
+    selectKey = \case
```
turion, line 40: Why not pattern match directly in the function?

rhine-terminal/TerminalSimple.hs
```diff
@@ -0,0 +1,100 @@
...
+import FRP.Rhine.Terminal (TerminalEventClock(..))
+import System.Exit (exitSuccess)
+
+-- Clocks
+
+data Input = Char Char
+           | Space | Backspace | Enter
+           | Exit
+
+type InputClock = SelectClock (TerminalEventClock LocalTerminal) Input
```
turion, line 34: In principle this clock could go in the library as well.

rhine-terminal/TerminalSimple.hs
```diff
@@ -0,0 +1,100 @@
...
+
+inputClock :: LocalTerminal -> InputClock
+inputClock term = SelectClock { mainClock = TerminalEventClock term, select = selectKey }
```
turion, line 37:
```suggestion
inputClock term = SelectClock
  { mainClock = TerminalEventClock term
  , select = selectKey
  }
```

rhine-terminal/TerminalSimple.hs
turion, line 40: Rename to `select`

rhine-terminal/tests/Main.hs
```diff
@@ -0,0 +1,69 @@
...
+
+import Test.Hspec
+import Data.Text (singleton)
+
+type KeyClock t = SelectClock (TerminalEventClock t) Char
+
+keyClock :: Terminal t => t -> KeyClock t
+keyClock term = SelectClock { mainClock = TerminalEventClock term, select = selectChar }
+  where
+    -- selectChar :: Terminal t => Tag (TerminalEventClock t) -> Maybe Char
```
turion, line 25:
```suggestion
    selectChar :: Terminal t => Tag (TerminalEventClock t) -> Maybe Char
```

rhine-terminal/tests/Main.hs
```diff
@@ -0,0 +1,69 @@
...
+    selectChar = \case
+      Right (KeyEvent (CharKey k) _) -> Just k
+      _ -> Nothing
```
turion, line 28:
```suggestion
    selectChar (Right (KeyEvent (CharKey k) _)) = Just k
    selectChar _ = Nothing
```

rhine-terminal/tests/Main.hs
```diff
@@ -0,0 +1,69 @@
...
+
+defaultSettings :: TQueue Event -> VirtualTerminalSettings
+defaultSettings eventQueue = VirtualTerminalSettings
+    { virtualType         = "xterm"
+    , virtualWindowSize   = pure (Size 3 10)
+    , virtualEvent        = readTQueue eventQueue
+    , virtualInterrupt    = retry
+    }
```
turion, line 36:
```suggestion
  { virtualType       = "xterm"
  , virtualWindowSize = pure (Size 3 10)
  , virtualEvent      = readTQueue eventQueue
  , virtualInterrupt  = retry
  }
```

rhine-terminal/tests/Main.hs
```diff
@@ -0,0 +1,69 @@
...
+
+displayDot :: (Terminal t, MonadScreen m) => ClSF m (KeyClock t) () ()
+displayDot = tagS >-> arrMCl (\_ -> do
+  putChar '.'
+  flush)
```
turion, line 41:
```suggestion
displayDot = constMCl $ do
  putChar '.'
  flush
```

rhine-terminal/tests/Main.hs
```diff
@@ -0,0 +1,69 @@
...
+      it "replaces virtual inputs by dots" $ do
+        eventQueue <- newTQueueIO
+        withVirtualTerminal (defaultSettings eventQueue) $ \t -> do
+          void $ liftIO $ forkIO $ runTerminalT (flow $ testRhine t) t
+          charEvent eventQueue t '1'
+          threadDelay $ 1000 * 1000
+          charEvent eventQueue t '2'
+          threadDelay $ 1000 * 1000
+          charEvent eventQueue t '3'
+          threadDelay $ 1000 * 1000
```
turion, line 63: Maybe we can reduce these delays a bit? Waiting 3 seconds in a test is quite a lot.

turion (changes requested):
Some small changes, mainly formatting and refactoring.

I'm still not entirely sure whether the `Terminal` value should reside in the clock, or in a `Reader` monad. What is the common way to do things:

1. Have one terminal and several clocks (including subclocks like `SelectClock`) connecting to it
2. Have several terminals and separate clocks for them

For 1., a `Reader` pattern might be more appropriate, for 2., the current code is better. Looking at the example, I'm getting the feeling that 1. is more common. But we'd have to investigate again whether this gives us any trouble with monads, classes and types.

turion:
> Oh I just realised there is a loose end still:
> 
> > the semigroup instance for TerminalEventClock seems wrong. But how can we write one?
> 
> ```haskell
> instance Terminal t => Semigroup (TerminalEventClock t) where
>   t <> _ = t
> ```
> 
> What does it mean and can we do better?

The reason for that is https://hackage.haskell.org/package/rhine-0.8.0.0/docs/FRP-Rhine-Clock-Select.html#v:schedSelectClocks which needs a `Semigroup` instance to schedule the clocks. This requirement will go away with #171.

We want to combine two terminals into one somehow, so if we have two `SelectClock (TerminalEventClock t)`, we could make a single terminal that emits all the events for the subclocks to select from. But that doesn't make much sense, especially not if we don't know the type of `t`.

For me this seems to be a further hint that we actually want the `ReaderT` pattern in https://github.com/turion/rhine/pull/165#pullrequestreview-1026834756. (Sorry that I initially suggested otherwise, but I didn't understand `terminal` well enough.)

jmatsushita:
> I'm still not entirely sure whether the Terminal value should reside in the clock, or in a Reader monad. 

Yes that makes sense. In fact I tried to go the Reader route earlier but got stuck because `TerminalT` although a newtype over `ReaderT` [doesn't export its constructor](https://github.com/lpeterse/haskell-terminal/blob/master/src/System/Terminal/TerminalT.hs#L4) nor has a `MonadReader` instance... Unless I'm mistaken, we can't work around this, so I'll submit a PR and we can depend on a fork for now.

rhine-terminal/TerminalSimple.hs
jmatsushita: I'm not so sure about this. The value returned by `awaitEvent` is [of type `Either Interrupt Event`](https://hackage.haskell.org/package/terminal-0.2.0.0/docs/System-Terminal.html#g:13) and I think it should be up to the user to decide to filter this source and which data types to map to. In this example I map onto a kitchen sink `Input` ADT  (which also maps any interrupt to `Exit`, while in the more [involved example](https://github.com/turion/rhine/pull/175/files#diff-86197bb5ec43f3df9e0f77886bf50cbc3cbc651bb977058389cd29e970b2894dR46) interrupts have their own `SignalClock`.

rhine-terminal/TerminalSimple.hs
jmatsushita: Sounds good. Also added `NamedFieldPuns` to make it more terse.

jmatsushita: It makes it a bit more verbose with extra parens and duplication of  `select` but I still made the change:
```haskell
select (Right (KeyEvent (CharKey k) m))  = Just (Char k m)
select (Right (KeyEvent SpaceKey _))     = Just Space
select (Right (KeyEvent BackspaceKey _)) = Just Backspace
select (Right (KeyEvent EnterKey _))     = Just Enter
select (Left _)                          = Just Exit
select _                                 = Nothing
```

vs 

```haskell
select = \case
  Right (KeyEvent (CharKey k) _)  -> Just (Char k)
  Right (KeyEvent SpaceKey _)     -> Just Space
  Right (KeyEvent BackspaceKey _) -> Just Backspace
  Right (KeyEvent EnterKey _)     -> Just Enter
  Left _                          -> Just Exit
  _                               -> Nothing
```

rhine-terminal/TerminalSimple.hs
jmatsushita: There was some meaning to the groupings, but I'm not strongly attached to it.

rhine-terminal/TerminalSimple.hs
jmatsushita: Sounds good 👍

rhine-terminal/tests/Main.hs
jmatsushita: Sounds good. I used 200 ms per delay.

jmatsushita:
> Unless I'm mistaken, we can't work around this, so I'll submit a PR and we can depend on a fork for now.

I was mistaken. I could write the instance without needing the `TerminalT` constructor.

rhine-terminal/ChangeLog.md
```diff
@@ -0,0 +1,5 @@
+# Revision history for rhine-terminal
+
+## 0.7.1.0 -- 2022.05.21
```
turion, line 3:
```suggestion
## 0.8.0.1 -- 2022-05-21
```

rhine-terminal/TerminalSimple.hs
```diff
@@ -0,0 +1,111 @@
...
+promptSource :: ClSF App PromptClock () Text
+promptSource = flip T.cons " > " . (cycle " ." !!) <$> count
+
+inputSink ::  ClSF App cl Input ()
+inputSink = arrMCl $ \input -> do
+  case input of
+    -- Don't display Ctrl-J https://github.com/lpeterse/haskell-terminal/issues/17
+    Char c m
+      | c /= 'J' && m /= ctrlKey -> putChar c >> flush
+      | otherwise -> pure ()
```
turion, line 74: Refactor using `unless` or `when`

rhine-terminal/rhine-terminal.cabal
```diff
@@ -0,0 +1,72 @@
+-- Initial rhine-gloss.cabal generated by cabal init.  For further
+-- documentation, see http://haskell.org/cabal/users-guide/
+
+name:                rhine-terminal
+version:             0.8.0
```
turion, line 5:
```suggestion
version:             0.8.0.1
```

rhine-terminal/rhine-terminal.cabal
```diff
@@ -0,0 +1,72 @@
...
+cabal-version:       1.18
+
+source-repository head
+  type:     git
+  location: git@github.com:turion/rhine.git
+
+source-repository this
+  type:     git
+  location: git@github.com:turion/rhine.git
+  tag:      v0.8.0.0
```
turion, line 27:
```suggestion
  tag:      v0.8.0.1
```

rhine-terminal/rhine-terminal.cabal
```diff
@@ -0,0 +1,72 @@
...
+  build-depends:       base         >= 4.11 && < 4.16
+                     , exceptions   >= 0.10.4
+                     , transformers >= 0.5
+                     , rhine        == 0.8.0.0
+                     , dunai        >= 0.6
+                     , terminal     >= 0.2.0.0
+                     , time         >= 1.9.3
+  hs-source-dirs:      src
+  default-language:    Haskell2010
+  ghc-options:         -Wall
```
turion, line 41: Can you have a look at https://github.com/turion/rhine/commit/c3ef9c6df7566a1eee2006eb9dab98f12323bff7 and adapt this accordingly?

rhine-terminal/src/FRP/Rhine/Terminal.hs
```diff
@@ -0,0 +1,103 @@
+{- | Wrapper to write @terminal@ applications in Rhine, using concurrency.
+-}
+
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE UndecidableInstances #-}
```
turion, line 9: Oh no, why is this needed?

rhine-terminal/src/FRP/Rhine/Terminal.hs
```diff
@@ -0,0 +1,103 @@
...
+-- transformers
+import Control.Monad.Trans.Reader
+
+-- rhine
+import FRP.Rhine.Clock.Proxy ()
+import FRP.Rhine
+import Control.Monad.Trans.Class (lift)
+
+-- * Terminal clock
+data TerminalEventClock = TerminalEventClock
```
turion, line 41:
```suggestion
-- | A clock that ticks whenever events or interrupts on the terminal arrive.
data TerminalEventClock = TerminalEventClock
```

rhine-terminal/src/FRP/Rhine/Terminal.hs
```diff
@@ -0,0 +1,103 @@
...
+        )
+
+instance GetClockProxy TerminalEventClock
+
+instance Semigroup TerminalEventClock where
+  t <> _ = t
+
+-- type TerminalClSF t m = ClSF (TerminalT t m) TerminalEventClock () (Either Interrupt Event)
+
+flowTerminal
```
turion, line 65: A little haddock here?

rhine-terminal/src/FRP/Rhine/Terminal.hs
```diff
@@ -0,0 +1,103 @@
...
+  :: ( Terminal t
+     , Clock (TerminalT t IO) cl
+     , GetClockProxy cl
+     , Time cl ~ Time (In  cl)
+     , Time cl ~ Time (Out cl)
+     )
+  => t
+  -> Rhine (TerminalT t IO) cl () ()
```
turion, line 73: Shouldn't this work for arbitrary inner monads other than `IO`?

rhine-terminal/src/FRP/Rhine/Terminal.hs
```diff
@@ -0,0 +1,103 @@
...
+
+-- rhine
+import FRP.Rhine.Clock.Proxy ()
+import FRP.Rhine
+import Control.Monad.Trans.Class (lift)
+
+-- * Terminal clock
+data TerminalEventClock = TerminalEventClock
+
+instance (Terminal t) => Clock (TerminalT t IO) TerminalEventClock
```
turion, line 43:
```suggestion
instance (Terminal t, MonadIO m) => Clock (TerminalT t m) TerminalEventClock
```

turion: I'm running in circles trying to understand. Can we not do `(MonadInput m, MonadIO m) => Clock m TerminalEventClock`?

turion (changes requested):

rhine-terminal/TerminalSimple.hs
turion: Ok, good point. Then this serves as a good example how users can customize the type returned by a clock.

rhine-terminal/TerminalSimple.hs
turion: True. I'm starting to like your style here :)

rhine-terminal/TerminalSimple.hs
turion: Ah I see. If the grouping is important, then a separate type would be in order:

```haskell
data Whitespace
  = Space
  | Backspace
  | Enter
data Input
  = Char Char
  | Whitespace Whitespace
  | Exit
```

rhine-terminal/rhine-terminal.cabal
jmatsushita: Should I update all tags in `source-repository` and versions in all rhine packages?

rhine-terminal/src/FRP/Rhine/Terminal.hs
jmatsushita: Must have been previous flailing, removing it.

rhine-terminal/src/FRP/Rhine/Terminal.hs
jmatsushita: Ah yes that makes a lot of sense. We only need `MonadInput` to `awaitEvent`.

jmatsushita:
> > Unless I'm mistaken, we can't work around this, so I'll submit a PR and we can depend on a fork for now.

> I was mistaken. I could write the instance without needing the TerminalT constructor.

I'm stuck again.

I can't use:
```
mainRhine = inputRhine ||@ concurrently @|| promptRhine
```

```
• Couldn't match type ‘IO’ with ‘TerminalT LocalTerminal IO’
  Expected: Schedule App InputClock PromptClock
    Actual: Schedule IO InputClock PromptClock
```

So it seems I need a `terminalConcurrently` but I'm stuck as I want to try avoiding importing the `TerminalT` constructor to be able to [`ask` for the `Terminal t` value](https://github.com/turion/rhine/pull/165/files#diff-0775481f28932846cb0dcbe66b336b96536c9db6e8a87c226caf50c7224b6ea2R96) inside the `TerminalT (ReaderT t m a)`.

But then [I don't know how to write `terminalConcurrently` without it](https://github.com/turion/rhine/pull/165/files#diff-0775481f28932846cb0dcbe66b336b96536c9db6e8a87c226caf50c7224b6ea2R89-R92).

Also I think this is unrelated, but just wanted to check that [instead of using a concrete `TerminalT t m` (here still `IO`)](https://github.com/turion/rhine/pull/165/files#diff-0775481f28932846cb0dcbe66b336b96536c9db6e8a87c226caf50c7224b6ea2R69) I should also probably use a `MonadTerminal m` constraint in `flowTerminal`.

rhine-terminal/src/FRP/Rhine/Terminal.hs
jmatsushita: Yes, I'll transform this into a `MonadTerminal m` constraint when I figure out how to write `terminalConcurrently`

rhine-terminal/rhine-terminal.cabal
turion: Ah sorry, no, I still have the release branch for 0.8.0.1 locally, let me merge that first and then we can rebase.

turion:
This is a tough one. I guess either opening an upstream PR making the constructor available and using a fork until then, or using `unsafeCoerce` to replace the constructor (not sure this works though).

turion:
Note: The problem doesn't go away after #171 because there we also need to construct a `TerminalT`.

turion:
I tried the unsafeCoerce version. It's ugly, but I don't know whether you'd be willing to maintain a fork otherwise.

rhine-terminal/rhine-terminal.cabal
jmatsushita: Sounds good 👍

rhine-terminal/src/FRP/Rhine/Terminal.hs
jmatsushita: I could write `flowTerminal` such that it returned an `m` with the `MonadIO m` constraint, but when I write it with the `MonadTerminal m` constraint:

```

flowTerminal'
  :: ( MonadIO m2
     , MonadMask m2
     , Terminal t
     , MonadTerminal m1
     , m1 ~ TerminalT t m2
     , Clock m1 cl
     , GetClockProxy cl
     , Time cl ~ Time (In  cl)
     , Time cl ~ Time (Out cl)
     )
  => t
  -> Rhine m1 cl () ()
  -> m2 ()
flowTerminal' term clsf = flip runTerminalT term $ flow clsf
```

It's not clear whether I'm gaining much and the `m1 ~ Terminal t m2` constraint seems suspect and I feel like I'm going in circles too :D

rhine-terminal/src/FRP/Rhine/Terminal.hs
jmatsushita: Done!

rhine-terminal/rhine-terminal.cabal
jmatsushita: Ok I'll stand by on this one.

jmatsushita:
Ok I believe I've addressed all your comments.

rhine-terminal/TerminalSimple.hs
```diff
@@ -0,0 +1,113 @@
...
+
+type InputClock = SelectClock TerminalEventClock Input
+
+inputClock :: InputClock
+inputClock = SelectClock
+  { mainClock = TerminalEventClock
+  , select }
+    where
+      select :: Tag TerminalEventClock -> Maybe Input
+      select = \case
```
turion, line 50: In fact if you use `\case`, you can define it directly without a `where` clause! I like that a lot. I'll use `\case` more in my own code I think.

rhine-terminal/TerminalSimple.hs
```diff
@@ -0,0 +1,113 @@
...
+inputSource :: ClSF App InputClock () Input
+inputSource = tagS
+
+promptSource :: ClSF App PromptClock () Text
+promptSource = flip T.cons " > " . (cycle " ." !!) <$> count
+
+inputSink ::  ClSF App cl Input ()
+inputSink = arrMCl $ \case
+  -- Don't display Ctrl-J https://github.com/lpeterse/haskell-terminal/issues/17
+  Char c m  -> when (c /= 'J' && m /= ctrlKey) $ putChar c >> flush
```
turion, line 73:
```suggestion
  Char c m  -> when (c /= 'J' || m /= ctrlKey) $ putChar c >> flush
```
I think..?

turion, line 73:
```suggestion
  Char 'J' Ctrl  -> flush
  Char c m  -> putChar c >> flush
```
I think this is the clearest and most obviously correct solution.

rhine-terminal/src/FRP/Rhine/Terminal.hs
```diff
@@ -0,0 +1,118 @@
...
+import Prelude hiding (putChar)
+import Control.Concurrent ()
+import Control.Concurrent.MVar ()
+import Data.IORef ()
+import Unsafe.Coerce (unsafeCoerce)
+
+-- time
+import Data.Time.Clock ( getCurrentTime )
+
+-- terminal
```
turion, line 22:
```suggestion
```

rhine-terminal/src/FRP/Rhine/Terminal.hs
```diff
@@ -0,0 +1,118 @@
...
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE RecordWildCards #-}
+module FRP.Rhine.Terminal where
+
+-- base
+import Prelude hiding (putChar)
+import Control.Concurrent ()
+import Control.Concurrent.MVar ()
+import Data.IORef ()
```
turion, line 16: Are these unneeded? If yes, remove.

rhine-terminal/src/FRP/Rhine/Terminal.hs
```diff
@@ -0,0 +1,118 @@
...
+-- terminal
+-- terminal
+import System.Terminal ( awaitEvent, runTerminalT, Event, Interrupt, TerminalT, MonadInput )
+import System.Terminal.Internal ( Terminal )
+
+-- transformers
+import Control.Monad.Trans.Reader
+
+-- rhine
+import FRP.Rhine.Clock.Proxy ()
```
turion, line 31: Unneeded import?

rhine-terminal/src/FRP/Rhine/Terminal.hs
```diff
@@ -0,0 +1,118 @@
...
+import FRP.Rhine
+import Control.Monad.Trans.Class (lift)
+import Control.Monad.Catch (MonadMask)
```
turion, line 34: Sort these into the right import blocks

rhine-terminal/src/FRP/Rhine/Terminal.hs
```diff
@@ -0,0 +1,118 @@
...
+-- @
+--
+-- mainRhine :: MonadIO m => Rhine (TerminalT LocalTerminal m) TerminalEventClock () ()
+-- mainRhine = tagS >-> arrMCl (liftIO . print) @@ TerminalEventClock
+--
+-- main :: IO ()
+-- main = withTerminal $ \term -> `flowTerminal` term mainRhine
+--
+-- @
+
```
turion, line 73: Do haddocks work if there is an empty line here? Can you have a look?

rhine-terminal/src/FRP/Rhine/Terminal.hs
turion: Looking great :)

rhine-terminal/src/FRP/Rhine/Terminal.hs
```diff
@@ -0,0 +1,118 @@
...
+     , Time cl1 ~ Time cl2
+     )
+  => Schedule (TerminalT t IO) cl1 cl2
+terminalConcurrently
+  = Schedule $ \cl1 cl2 -> do
+      term <- terminalT ask
+      lift $ first liftTransS <$>
+        initSchedule concurrently (runTerminalClock term cl1) (runTerminalClock term cl2)
+
+-- Workaround TerminalT constructor not being exported. Should be safe in practice.
```
turion, line 104: Can you open an upstream issue and link to that?

rhine-terminal/src/FRP/Rhine/Terminal.hs
```diff
@@ -0,0 +1,118 @@
...
+
+runTerminalClock
+  :: Terminal t
+  => t
+  -> cl
+  -> RunTerminalClock IO t cl
+runTerminalClock term unhoistedClock = HoistClock
+  { monadMorphism = flip runTerminalT term
+  , ..
+  }
```
turion, line 118: Are these internal definitions or are they relevant for library users?

rhine-terminal/src/FRP/Rhine/Terminal.hs
```diff
@@ -0,0 +1,118 @@
+{- | Wrapper to write @terminal@ applications in Rhine, using concurrency.
+-}
+
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE RecordWildCards #-}
+module FRP.Rhine.Terminal where
```
turion, line 10: I've goaded you into deleting the export list here. Maybe this was premature. There are things like `terminalT` that we certainly don't want to export. There are two ways to achieve that. One is an export list. The other one is a module `FRP.Rhine.Terminal.Internal` that contains all definitions that should not be exported. I'm fine with both variants, but we should pick one to hide at least `terminalT`.

rhine-terminal/src/FRP/Rhine/Terminal.hs
turion: If internal they should not be exported, if relevant to library users they should be documented.

rhine-terminal/tests/Main.hs
```diff
@@ -0,0 +1,76 @@
...
+
+charEvent :: Terminal t => TQueue Event -> t -> Char -> IO ()
+charEvent eventQueue terminal char = do
+  termCommand terminal $ PutText $ singleton char
+  atomically $ writeTQueue eventQueue $ KeyEvent (CharKey char) mempty
+
+main :: IO ()
+main = hspec $ do
+  describe "rhine-terminal with VirtualTerminal" $ do
+    it "replaces virtual inputs by dots" $ do
```
turion, line 61: This description isn't quite true anymore.

turion (changes requested):
I hope you'll bear with my meticulous review style :sweat_smile: I can find a lot of small details if I go looking.

rhine-terminal/TerminalSimple.hs
```diff
@@ -0,0 +1,113 @@
+{- | Example application for the @rhine-terminal@ library. -}
+
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE UndecidableInstances #-}
```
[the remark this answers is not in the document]
jmatsushita: Since I use `Text` OverloadedString is nice. The rest is not used indeed,

rhine-terminal/TerminalSimple.hs
jmatsushita: Oops, yes indeed. I've also moved this to the clock/source definition as a guard, as we [can't use patterns](https://github.com/lpeterse/haskell-terminal/issues/14).

jmatsushita: Unfortunately, we [can't use patterns](https://github.com/lpeterse/haskell-terminal/issues/14) and `Modifiers` are not exported, so we can't just pattern match :/ I'll probably add this to my fork at some point...

rhine-terminal/src/FRP/Rhine/Terminal.hs
jmatsushita: They work but add a newline, so I've tightened it up.

rhine-terminal/src/FRP/Rhine/Terminal.hs
jmatsushita: I linked to the PR I opened https://github.com/lpeterse/haskell-terminal/pull/18

rhine-terminal/src/FRP/Rhine/Terminal.hs
jmatsushita: I opted for the explicit export list 👍

rhine-terminal/tests/Main.hs
jmatsushita: Oh, indeed I've changed the test to go back to the original idea.

jmatsushita:
> I hope you'll bear with my meticulous review style 😅 I can find a lot of small details if I go looking.

Not at all! Thanks for your patience, I didn't have tons of time in the past couple of weeks :) I think I've addressed all your comments. Feel free to request more changes if you catch anything 👍

turion (approved):
Fantastic! Thanks a lot for contributing!

PR #171, opened 2022-06-25
Dev monad schedule without diff

turion:
See https://hackage.haskell.org/package/monad-schedule. By using this package, I hope to simplify

* type signatures: the monad doesn't appear in the clock type of `SequentialClock` and `Parallelclock` anymore
* schedule implementations: these aren't needed in most cases anymore because most monads will have a `MonadSchedule` instance
* design choices: no need to choose a schedule
* Rhine concepts: only 3 concepts (clocks, ClSFs, SNs) to memorise

The development in this branch is fundamental to other improvements like https://github.com/turion/rhine/pull/143.

## To do

* [ ] Changelog entry
* [x] Update cheatsheet
* [x] "Transition from article" page
* [x] Fix https://github.com/turion/monad-schedule/pull/10 first, release, and depend on that version
* [x] Add unit tests for pure clocks and see whether these need some scheduling monad (e.g. `Yield` at least)
* [x] Merge https://github.com/turion/rhine/pull/233 first
* [ ] Bump version to 1.0 I guess :)
* [x] Run all executables


turion:
Successful pipeline run depends on merging https://github.com/commercialhaskell/stackage/pull/6734 and a further nixpkgs update. Or else patching the version in the flake temporarily.

rhine/src/FRP/Rhine/Schedule.hs
```diff
@@ -20,117 +20,70 @@ Specific implementations of schedules are found in submodules.
...
 module FRP.Rhine.Schedule where
 
--- transformers
-import Control.Monad.Trans.Reader
+-- base
+import Data.List.NonEmpty (NonEmpty (..))
+import qualified Data.List.NonEmpty as N
 
 -- dunai
+
```
turion, line 28:
```suggestion
```

rhine-terminal/src/FRP/Rhine/Terminal.hs
```diff
@@ -84,38 +82,13 @@ flowTerminal ::
...
-runTerminalClock ::
-  Terminal t =>
-  t ->
-  cl ->
-  RunTerminalClock IO t cl
-runTerminalClock term unhoistedClock =
-  HoistClock
-    { monadMorphism = flip runTerminalT term
-    , ..
-    }
```
turion, line 121: Why delete this?

turion: Because it was not part of the public interface, and only used in the schedule. Nevertheless I imagine it could be useful in situations where one doesn't want to lift everything to `TerminalT`.

turion:
It seems that CI is not stable enough for millisecond resolution tests. I will have to decrease accuracy to 10th of a millisecond or so. Or seconds right away?

turion:
Changelog and version bump are for the release PR. I'll do a 0.8 release before the 1.0 release first though.

rhine-gloss/src/FRP/Rhine/Gloss/Pure.hs
```diff
@@ -28,30 +27,38 @@ import qualified Control.Category as Category
...
 
 -- rhine
 import FRP.Rhine
-import FRP.Rhine.Reactimation.ClockErasure
+
+-- monad-schedule
+import Control.Monad.Schedule.Class
+import Control.Monad.Schedule.Yield
+import Control.Monad.Trans.MSF (performOnFirstSample)
+import Data.Functor.Identity
```
turion, line 44: Sort imports

PR #181, opened 2022-08-08
Add a loop or feedback SN constructor

turion:
* [x] Test at least with an example application
* [x] Consider renaming `Loop` to `Feedback`
* [x] Resolve FIXMEs
* [x] Add missing cases in combinators (this should have been a compile error, separate PR)

Thanks @SheetKey for inspiration

### DIFF
--- a/rhine-examples/rhine-examples.cabal
+++ b/rhine-examples/rhine-examples.cabal
@@ -77,6 +77,18 @@ executable Sawtooth
   if flag(dev)
     ghc-options: -Werror
 
+executable RandomWalk
+  hs-source-dirs:      src
+  main-is:             RandomWalk.hs
+  ghc-options:         -Wall -threaded -rtsopts -with-rtsopts=-N
+  build-depends:       base         >= 4.9 && < 5
+                     , rhine        == 0.8.0.1
+                     , random       >= 1.1
+                     , simple-affine-space
+  default-language:    Haskell2010
+  if flag(dev)
+    ghc-options: -Werror
+
 flag dev
   description: Enable warnings as errors. Active on ci.
   default: False
--- a/rhine-examples/src/RandomWalk.hs
+++ b/rhine-examples/src/RandomWalk.hs
@@ -1,0 +1,82 @@
+{- | Simulation of a random walk.
+
+The internal state is a point in 2D space.
+Every millisecond, a unit step is taken in a random direction along either the X or Y axis.
+The current position and the distance to the origin is shown, as well as the position and distance to a saved point.
+(A point can be saved by pressing enter.)
+
+This mainly exists to test the 'feedbackRhine' construct.
+-}
+
+{-# LANGUAGE Arrows #-}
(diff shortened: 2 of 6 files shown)


turion:
@SheetKey can you have a look at the `RandomWalk` example and see whether the function `feedbackRhine` and the `Loop` constructor cover your use cases?

turion:
* [ ] Merge https://github.com/turion/rhine/pull/182 first

turion:
* [ ] Clean up git history
* [ ] Check & complete haddocks

SheetKey:
@turion its exactly what I was looking for! I'm using an ECS and using only ClSFs with feedback for my main game loop forced everything to happen on the same clock where I would rather have the ability to have different processes happen at different rates.

turion:
That sounds great :) what's an ECS? In case your code is open, feel free to share.

SheetKey:
Entity component system. You essentially treat every game object as a collection of components (mostly wrapped in Maybe), then define functions, in my case signal functions, that do things with the components. For example I have [position](https://github.com/SheetKey/LearnRhine/blob/63da39232da9a860a7d841b9ea886eeaab4f3ab7/src/FRP/Rhine/SDL/Components/Position.hs#L44) and [velocity](https://github.com/SheetKey/LearnRhine/blob/63da39232da9a860a7d841b9ea886eeaab4f3ab7/src/FRP/Rhine/SDL/Components/Velocity.hs#L8) components with corresponding signal functions that [update the velocity](https://github.com/SheetKey/LearnRhine/blob/63da39232da9a860a7d841b9ea886eeaab4f3ab7/src/FRP/Rhine/SDL/Process/Velocity.hs#L57) (from user input in the players case) and another that [updates position](https://github.com/SheetKey/LearnRhine/blob/63da39232da9a860a7d841b9ea886eeaab4f3ab7/src/FRP/Rhine/SDL/Process/Movement.hs#L48) from the velocity. My game state is a list of entities and I compose signal functions that do various things. Right not my ECS is just a [big record](https://github.com/SheetKey/LearnRhine/blob/63da39232da9a860a7d841b9ea886eeaab4f3ab7/src/FRP/Rhine/SDL/Entity.hs#L45) but there are some libraries that I'm considering using once I have a better grasp of SDL: [apecs](https://hackage.haskell.org/package/apecs) and [ecstasy](https://hackage.haskell.org/package/ecstasy).

turion:
Ah yes :) that's the kind of thing that we started to develop dunai & rhine for. Have you looked into `gloss` and `rhine-gloss`? In case you want some simple SDL 2D vector graphics without worrying too much about the backend.

SheetKey:
I haven't really looked into gloss. I decided on sdl2 because of the various graphics libraries for haskell it seemed like the most well documented for game programming. Even though its mostly C++ it translates to haskell with very little changes. That being said I haven't done any vector graphics, just loading pngs. If I run into issues I might switch but I haven't had any problems using sdl so far.

PR #188, opened 2022-09-23
Add forecastWith

turion:


reubenharry:
Ah, very cool, I think this will be useful! What do the a1, a2... represent here? I'm not quite sure I understand exactly what the function does.

turion:
Since it's a signal _function_, it will need some input. You need to supply the forecasted input as well. For a pure signal, you'll have `a = ()`, so you can supply `[((), ts1), ((), ts2), ...]`.

turion:
This doesn't build because stack and nixpkgs don't know about the new version of `time-domain` yet.

* [ ] Increase `time-domain` version bound
* [ ] Either wait until stack & nixpkgs are updated or overwrite versions

PR #194, opened 2022-12-13
Run cabal update on CI

turion:


turion:
The issue is that dunai doesn't build with `simple-affine-space == 0.2`, but `dunai == 0.8.3` didn't include the upper version bound that prevented this version.

turion:
Also see https://github.com/ivanperez-keera/dunai/issues/298

turion:
The upper version bound was introduced in `dunai == 0.9.0`, together with `MonadRandom == 0.6`, which is the reason why `dunai >= 0.9` isn't chosen.

turion:
The cabal file of `MonadRandom` has been revised (https://github.com/byorgey/MonadRandom/issues/51#issuecomment-1356863694), let's try again.

PR #222, opened 2023-05-12
Effective Sample Size & tempering/annealing

turion:
Right now, the best approach to estimating constant parameters seems to be a stochastic diffusion process that keeps & mutates guesses for these parameters in its state, and lowers its temperature with time. This might be called "tempering" or "annealing" in the literature. There are two open research questions:

* [ ] How should the diffusion rate change? It should probably decrease with time, but also increase when the effective sample size is low. How exactly is maybe a matter of taste
* [ ] How does diffusion work for a generic parameter? One might want to make a random walk/MCMC-step on a prior (or posterior) distribution, then this seems to be RMSMC. But the implementation of `mhStep` in `monad-bayes` seems questionable. The other idea would be deriving a diffusion kernel from the parameter type.

### DIFF
--- a/rhine-bayes/app/Main.hs
+++ b/rhine-bayes/app/Main.hs
@@ -235,6 +235,7 @@ mains =
   [ ("single rate", mainSingleRate)
   , ("multi rate, temperature process", mainMultiRate)
   , ("multi rate, temperature process, RMSMC", mainMultiRateRMSMC)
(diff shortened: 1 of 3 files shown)


turion:
> when the effective sample size is low

To make this a more relevant criterion, one should first take the whole population `PopulationT m everything` and project it onto only that parameter, collapsing particles with identical values to one, and then looking at the ESS. But that would require an `Eq` constraint.

PR #223, opened 2023-05-12
RMSMC

turion:
Trying to implement Resample-Move Sequential Monte Carlo.


turion:
@idontgetoutmuch I'm having performance issues here. Reproduce with `cd rhine-bayes && cabal run rhine-bayes-gloss`.

PR #244, opened 2023-08-07
Dev ppmmh

turion:
@reubenharry this is my take on PMMH. So far it's single-threaded and synchronous. The difference to standard PMMH is that the parameter chain does not quite follow MCMC, but SMC.

Right now it doesn't work very well yet, the variance is quite big.


turion:
I might have just fixed it, the variance looks a lot better now. The issue was again that I forgot that after a resampling step, all particles are already normalized.

PR #249, opened 2023-09-15
bump dunai to 0.11

ners:
This also required a flake update to bring in simple-affine-space-0.2.

PR #251, opened 2023-10-01
Add support for dunai 0.11

blackheaven:
I'm a bit struggling with instances:

```
src/FRP/Rhine/ClSF/Util.hs:338:19: error:
    • Could not deduce (Floating
                          (MSF (Control.Monad.Trans.Reader.ReaderT (TimeInfo cl) m) v s))
        arising from a use of ‘^-^’
      from the context: (Monad m, Eq s, Floating s, VectorSpace v s,
                         s ~ Diff td, td ~ Time cl)
        bound by the type signature for:
                   highPass :: forall (m :: Type -> Type) s v td.
                               (Monad m, Eq s, Floating s, VectorSpace v s, s ~ Diff td) =>
                               Diff td -> BehaviourF m td v v
        at src/FRP/Rhine/ClSF/Util.hs:(328,1)-(337,21)
    • In the expression: clId ^-^ lowPass t
      In an equation for ‘highPass’: highPass t = clId ^-^ lowPass t
    |
338 | highPass t = clId ^-^ lowPass t
    |                   ^^^
```

which is weird since:

```
instance (Monad m, Eq s, Num s, VectorSpace v s, Floating (MSF m a s)) => VectorSpace (MSF m a v) (MSF m a s)
instance (Monad m, Floating b) => Floating (MSF m a b)
```

If anyone can have a look

### DIFF
--- a/rhine-bayes/rhine-bayes.cabal
+++ b/rhine-bayes/rhine-bayes.cabal
@@ -33,7 +33,7 @@ library
   build-depends:       base         >= 4.11 && < 4.18
                      , transformers >= 0.5
                      , rhine        == 1.0
-                     , dunai        ^>= 0.9
+                     , dunai        >= 0.9 && < 0.12
                      , log-domain   >= 0.12
                      , monad-bayes  ^>= 1.1
   hs-source-dirs:      src
--- a/flake.lock
+++ b/flake.lock
@@ -3,11 +3,11 @@
     "flake-compat": {
       "flake": false,
       "locked": {
-        "lastModified": 1650374568,
-        "narHash": "sha256-Z+s0J8/r907g149rllvwhb4pKi8Wam5ij0st8PwAh+E=",
+        "lastModified": 1673956053,
+        "narHash": "sha256-4gtG9iQuiKITOjNQQeQIpoIB6b16fm+504Ch3sNKLd8=",
         "owner": "edolstra",
         "repo": "flake-compat",
-        "rev": "b4a34015c698c7793d592d66adbab377907a2be8",
+        "rev": "35bb57c0c8d8b62bbfd284272c928ceb64ddbde9",
         "type": "github"
       },
       "original": {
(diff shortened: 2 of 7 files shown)


rhine-bayes/rhine-bayes.cabal
```diff
@@ -30,10 +30,10 @@ library
     FRP.Rhine.Bayes
   other-modules:
     Data.MonadicStreamFunction.Bayes
-  build-depends:       base         >= 4.11 && < 4.18
```
turion, line 33: I'd rather keep tight version bounds on `base`, so one should bump to 4.19 here.

turion:
> I'm a bit struggling with instances:

Ah yes, that one is really tricky, it took me some time. You need:
```
import Data.MonadicStreamFunction.Instances.Num ()
```

turion:
I think you're actually bumping the dunai version, and not so much the GHC version, right?

blackheaven:
> Ah yes, that one is really tricky, it took me some time. You need:

Thank you so much!

blackheaven:
> I think you're actually bumping the dunai version, and not so much the GHC version, right?

and `simple-affine-space`, but that's the only blockers to GHC 9.4 support.

turion:
But GHC 9.4 is already supported in `rhine`. I think to support _9.6_, one needs to bump `dunai`, but I'm not sure. The issue with 9.4 is only that the dependency `monad-bayes` didn't support 9.4 for a long time. But it does now, I think we just need to enable the CI jobs again.

blackheaven:
> But GHC 9.4 is already supported in `rhine`. I think to support _9.6_, one needs to bump `dunai`, but I'm not sure. The issue with 9.4 is only that the dependency `monad-bayes` didn't support 9.4 for a long time. But it does now, I think we just need to enable the CI jobs again.

I see, when I tried, I have just updated `flake.lock` and I have got issues I have attributed to GHC 9.4, which in fact where packages set changes.

I have `monad-bayes` on my list, once done, I'll come back to finish the work :)

rhine/src/FRP/Rhine/ClSF/Core.hs
```diff
@@ -1,5 +1,6 @@
 {-# LANGUAGE Arrows #-}
 {-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
```
turion, line 3: This is not necessary, right?

turion:
Can you fix fourmolu and the remaining build errors?

blackheaven:
I had a fix, I forgot to push it :/

blackheaven:
Is there anything I can do to make the PR move forward?

turion:
Sorry, I overlooked that there was already a PR open to fix this in https://github.com/turion/rhine/pull/249 :( It was open since longer and I wasn't sure it would be merged. Thank you anyways!

PR #269, opened 2023-12-31
Change Stdin clock to text

turion:


turion:
I'm of a mind to either drop the stack setup completely or to drop GHC < 9.4. In the latter case, I should bump to GHC 9.8 first.

PR #283, opened 2024-01-30
Fix MacOS runtime error in Gloss apps

turion:
* [ ] Test whether this is in fact the relevant change (@reubenharry can you help maybe?)
* [ ] Document in the `rhine-gloss` readme
* [ ] Port to other gloss example
* [ ] Open analogous tickets for `sonnendemo` and eolc


turion:
But note that not having `-threaded` may lead to worse performance on Macs, as then all GHC threads are blocked by OpenGL FFI calls.

PR #287, opened 2024-02-08
Dev catch

turion:
Introduces some form of control flow for clocks. Clocks may now throw exceptions, and other clocks can be started as exception handlers.

* [x] Clean up git history
* [x] Figure out whether there are ways to catch the clock exception without discarding the state of the clsf
* [x] Reduce redundancy between IO errors and `ExceptT`. 
  * [x] Merge only `ExceptT` versions here, open separate PR for IO versions, needs to add benchmarks (to be merged after #285 ). See https://github.com/turion/rhine/pull/300

PR #297, opened 2024-03-09
Add Never clock

ners:
As seen in: https://github.com/ners/rhine-linux/blob/master/kitchen-sink/Main.hs

Some clocks may fail to initialise (e.g. because they depend on missing external resources), but we would like to recover gracefully by replacing them with a clock that never ticks.

And so we introduce a clock that never ticks, but can produce anything:

```haskell
initFallibleClock
    :: forall m cl
     . ( Clock m cl
       , MonadIO m
       , MonadCatch m
       , Time cl ~ Time Never
       )
    => cl
    -> RunningClockInit m (Time cl) (Tag cl)
initFallibleClock cl = catchAll (initClock cl) $ const $ initClock Never <&> first (>>^ second absurd)
```

### DIFF
--- a/rhine/src/FRP/Rhine/Clock/Realtime/Never.hs
+++ b/rhine/src/FRP/Rhine/Clock/Realtime/Never.hs
@@ -1,0 +1,36 @@
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE TypeFamilies #-}
+
+-- | A clock that never ticks.
+module FRP.Rhine.Clock.Realtime.Never where
+
+-- base
+import Control.Concurrent (threadDelay)
+import Control.Monad (forever)
+import Data.Void (Void)
+
+-- time
+import Data.Time.Clock
+
+-- rhine
+import FRP.Rhine.Clock
+import FRP.Rhine.Clock.Proxy
+
+-- transformers
+import Control.Monad.IO.Class
+
+-- | A clock that never ticks.
+data Never = Never
+
+instance (MonadIO m) => Clock m Never where
+  type Time Never = UTCTime
+  type Tag Never = Void
+
+  initClock _ = do
+    initialTime <- liftIO getCurrentTime
+    return
+      ( constM (liftIO . forever . threadDelay $ 10 ^ 9)
+      , initialTime
+      )
+
+instance GetClockProxy Never
--- a/rhine/ChangeLog.md
+++ b/rhine/ChangeLog.md
@@ -1,5 +1,9 @@
 # Revision history for rhine
 
+## 1.2.1
+
+* Added `FRP.Rhine.Clock.Realtime.Never` (clock that never ticks)
+
 ## 1.2
 
 * Changed Stdin clock Tag type to Text
--- a/rhine/rhine.cabal
+++ b/rhine/rhine.cabal
@@ -86,6 +86,7 @@ library
     FRP.Rhine.Clock.Realtime.Busy
     FRP.Rhine.Clock.Realtime.Event
     FRP.Rhine.Clock.Realtime.Millisecond
+    FRP.Rhine.Clock.Realtime.Never
     FRP.Rhine.Clock.Realtime.Stdin
     FRP.Rhine.Clock.Select
     FRP.Rhine.Clock.Unschedule
--- a/rhine/src/FRP/Rhine.hs
+++ b/rhine/src/FRP/Rhine.hs
@@ -38,6 +38,7 @@ import FRP.Rhine.Clock.Realtime.Audio as X
 import FRP.Rhine.Clock.Realtime.Busy as X
 import FRP.Rhine.Clock.Realtime.Event as X
 import FRP.Rhine.Clock.Realtime.Millisecond as X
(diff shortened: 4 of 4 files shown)


turion:
Thanks a lot! This is a useful addition.

> This clock does not actually run in MonadIO because it is modelled after FRP.Rhine.Clock.Realtime.Busy, which runs in IO. We should probably make both of them MonadIO in the future.

I agree, we should, and indeed I think `Never` can already be in `MonadIO`. `Busy` can be changed later.

rhine/src/FRP/Rhine/Clock/Never.hs
```diff
@@ -0,0 +1,35 @@
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE TypeFamilies #-}
+
+-- | A clock that never ticks.
+module FRP.Rhine.Clock.Never where
```
turion, line 5:
```suggestion
module FRP.Rhine.Clock.Realtime.Never where
```

To fit better in the existing module hierarchy, I believe this clock should be in the `Realtime` namespace.

rhine/src/FRP/Rhine/Clock/Never.hs
```diff
@@ -0,0 +1,35 @@
...
+-- base
+import Data.Time.Clock
+
+-- rhine
+import FRP.Rhine.Clock
+import FRP.Rhine.Clock.Proxy
+
+import Control.Concurrent (threadDelay)
+import Control.Monad (forever)
+import Data.Kind (Type)
```
turion, line 16:
```suggestion
import Control.Concurrent (threadDelay)
import Control.Monad (forever)
import Data.Kind (Type)

-- time
import Data.Time.Clock

-- rhine
import FRP.Rhine.Clock
import FRP.Rhine.Clock.Proxy
```

rhine/src/FRP/Rhine/Clock/Never.hs
```diff
@@ -0,0 +1,35 @@
...
+
+{- |
+A clock that never ticks.
+-}
```
turion, line 20:
```suggestion
-- | A clock that never ticks.
```

rhine/src/FRP/Rhine/Clock/Never.hs
```diff
@@ -0,0 +1,35 @@
...
+
+instance Clock IO (Never tag) where
+  type Time (Never _) = UTCTime
+  type Tag (Never tag) = tag
+
+  initClock _ = do
+    initialTime <- getCurrentTime
+    return
+      ( constM getCurrentTime
+          &&& arrM (const . forever . threadDelay $ 10 ^ 9)
```
turion, line 31: You should be able to use `constM` here.

turion:
Have you seen https://github.com/turion/rhine/pull/287? I think this will interplay really nicely with `Never`!

rhine/src/FRP/Rhine/Clock/Never.hs
```diff
@@ -0,0 +1,35 @@
...
+import Control.Concurrent (threadDelay)
+import Control.Monad (forever)
+import Data.Kind (Type)
+
+{- |
+A clock that never ticks.
+-}
+data Never (tag :: Type) = Never
+
+instance Clock IO (Never tag) where
```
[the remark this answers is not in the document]
ners: Of course.

rhine/src/FRP/Rhine/Clock/Never.hs
ners: Excellent idea; obviously if `forever` can produce `tag`, it can also produce `(time, tag)`. :slightly_smiling_face:

rhine/src/FRP/Rhine/Clock/Never.hs
ners: Yup, makes sense since we're using `UTCTime`.

rhine/src/FRP/Rhine/Clock/Realtime/Never.hs
```diff
@@ -0,0 +1,36 @@
...
+
+-- transformers
+import Control.Monad.IO.Class
+
+-- | A clock that never ticks.
+data Never (tag :: Type) = Never
+
+instance (MonadIO m) => Clock m (Never tag) where
+  type Time (Never _) = UTCTime
+  type Tag (Never tag) = tag
```
turion, line 27:
```suggestion
data Never = Never

instance (MonadIO m) => Clock m (Never tag) where
  type Time Never = UTCTime
  type Tag Never = Void
```

I believe it should be possible not having to specify the tag at all. By using `Void`, it really proves that no tag is ever produced. This has the advantage that if e.g. you have a `ParallelClock Never cl`, then `Tag (ParallelClock Never cl) = Either (Tag Never) (Tag cl)`, so when you encounter a `Left tag` value, you can always prove its absence using `absurd`.

ners: Nice idea. In places where we `initClock` we can simply map the type with either `undefined` or `absurd`.

PR #302, opened 2024-03-26
Change Busy clock effect to MonadIO

ners:
This brings the Busy clock in line with the other Realtime clocks.

As discussed in https://github.com/turion/rhine/pull/297#issuecomment-1987961348


turion:
Thanks a lot!

PR #306, opened 2024-04-16
Add trivial clock

turion:


rhine/src/FRP/Rhine/Clock/Trivial.hs
```diff
@@ -0,0 +1,19 @@
...
+-- base
+import Control.Arrow
+
+-- rhine
+import FRP.Rhine.Clock
+import FRP.Rhine.Clock.Proxy (GetClockProxy)
+
+data Trivial = Trivial
+
+-- FIXME does this schedule right?
```
turion, line 13: It doesn't matter, the user can choose any schedule they like.

PR #299, opened 2024-03-11
Replace dunai by automata/state machines

turion:
# Motivation

Unfortunately there are two reasons that force me to drop the dependency on dunai, and replace it by an implementation of effectful Mealy state machines.

1. `dunai` is incompatible with transformers 0.6, mtl 2.3, GHC 9.6: https://github.com/turion/rhine/pull/215, https://github.com/ivanperez-keera/dunai/issues/402
2. There are fundamental performance issues with `dunai`: https://github.com/turion/rhine/issues/227 https://github.com/ivanperez-keera/dunai/issues/370

# Work done

This PR replaces dunai by a effectful Mealy state machines, in the initial encoding. The implementation is heavily inspired by https://github.com/lexi-lambda/incremental/blob/master/src/Incremental/Fast.hs and https://github.com/turion/essence-of-live-coding. It solves the two problems:

1. The automaton implementation is compatible with transformers 0.6. Prominently, there is no support for the old `ListT`.
2. A benchmark is added which shows dramatic performance improvements.

## Benchmark

The benchmark is a simple word count implementation which counts the words of Shakespeare's complete works. Its purpose is to show how much overhead `dunai` and `rhine` introduce. It includes:

* An idiomatic Rhine implementation
* An idiomatic state automaton implementation (without Rhine clocks)
* An idiomatic dunai implementation
* Three direct implementations using `text`
  * Using `IORef`s
  * Reading line by line, but avoiding `IORef`s
  * Lazy text

For a non-Haskell baseline, the standard `wc` completes the benchmark in 30 ms on my reference machine.

### Result with state automata

```
benchmarking WordCount/rhine
time                 239.8 ms   (233.2 ms .. 253.6 ms)
                     0.998 R²   (0.993 R² .. 1.000 R²)
mean                 214.1 ms   (196.9 ms .. 225.8 ms)
std dev              19.36 ms   (9.243 ms .. 27.70 ms)
variance introduced by outliers: 16% (moderately inflated)

benchmarking WordCount/automaton
time                 84.87 ms   (77.64 ms .. 92.24 ms)
                     0.992 R²   (0.986 R² .. 0.998 R²)
mean                 91.60 ms   (88.83 ms .. 94.73 ms)
std dev              5.279 ms   (3.818 ms .. 7.088 ms)
variance introduced by outliers: 18% (moderately inflated)

benchmarking WordCount/dunai
time                 900.6 ms   (789.3 ms .. 972.3 ms)
[...]

PR #309, opened 2024-05-10
Improve clock erasure in the light of automata

turion:
Follow-up of #299. Tries to improve how clock erasure works. Ideally, after clock erasure, `flow` should just be a call to `reactimate` on a concrete automaton.


turion:
Some variant of this has been merged long ago and the remaining ideas (strictness in OptimizedStreamT constructors) didn't seem to lead anywhere

PR #327, opened 2024-05-22
Dev gloss improvements

turion:


turion:
I don't really understand why I added the condition in https://github.com/turion/rhine/pull/254. I should document better the next time.

PR #339, opened 2024-06-25
Relax tasty-quickcheck bounds

turion:


turion:
Should solve https://github.com/commercialhaskell/stackage/issues/7460

PR #341, opened 2024-06-25
Rewrite Nix flake

ners:
As discussed at ZuriHac: the new and improved Rhine flake! ✨

Design goals:
 - minimise external dependencies; only Nixpkgs
 - use Nixpkgs to support all platforms and GHC versions
 - composability: expose an overlay that can be used by others

This is the sort of flake I wish to import when working on Rhine-powered projects. Until now I've had to copy around a bunch of boilerplate.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -21,6 +21,7 @@
       lib = inputs.nixpkgs.lib;
 
       # The names of all Haskell packages in this repository, defined as all the directories with *.cabal files in them.
+      # Contains e.g.: rhine, rhine-examples, rhine-bayes, ...
       pnames = map (path: baseNameOf (dirOf path)) (lib.fileset.toList (lib.fileset.fileFilter (file: file.hasExt "cabal") ./.));
 
       # All GHC versions that this project is tested with.
@@ -38,7 +39,7 @@
         // { default = pkgs.haskellPackages; };
 
       # A haskellPackages overlay containing everything defined in this repo
-      rhinePackages = hfinal: hprev:
+      rhinePackagesOverlay = hfinal: hprev:
         lib.genAttrs pnames (pname: hfinal.callCabal2nix pname ./${pname} { });
 
       # A nixpkgs overlay containing everything defined in this repo, for reuse in downstream projects
@@ -78,20 +79,22 @@
         {
           # The Haskell package set containing the packages defined in this repo
           haskell = prev.haskell // {
-            packageOverrides = with prev.haskell.lib.compose; lib.composeManyExtensions ([
+            packageOverrides = lib.composeManyExtensions ([
               prev.haskell.packageOverrides
-              rhinePackages
+              rhinePackagesOverlay
             ] ++ temporaryHaskellOverrides);
           };
 
           # Helper packages containing aspects of the whole rhine build:
           # All executables, built with the nixpkgs-default GHC
+          # We build these only with one GHC because otherwise the bin names would clash
           rhine-bin = prev.buildEnv
             {
               name = "rhine-bin";
               paths = map (pname: hps.default.${pname}) pnames;
               pathsToLink = [ "/bin" ];
             };
+
           # All libraries for all GHC versions
           rhine-lib = prev.buildEnv
             {
@@ -101,13 +104,15 @@
                 { hp = attrValues hps; pname = pnames; };
               pathsToLink = [ "/lib" ];
             };
-          # All haddocks
+
+          # Haddocks for all packages that can be uploaded to Hackage
           rhine-docs = prev.buildEnv
             {
               name = "rhine-docs";
               paths = map (pname: prev.haskell.lib.documentationTarball hps.default.${pname}) libPnames;
             };
-          # All sdist tarballs for Hackage publication
+
+          # Sdist tarballs for all packages that can be uploaded to Hackage
           rhine-sdist = prev.buildEnv
             {
               name = "rhine-sdist";
@@ -137,16 +142,21 @@
       # Usage: nix fmt
       formatter = forAllPlatforms (system: pkgs: pkgs.nixpkgs-fmt);
 
-      # Usage: nix build # This builds all rhine packages on all GHCs
+      # This builds all rhine packages on all GHCs, as well as docs and sdist
+      # Usage: nix build
       packages = forAllPlatforms (system: pkgs: {
         default = pkgs.rhine-all;
       });
 
-      legacyPackages = forAllPlatforms (system: pkgs: {
-        inherit (pkgs) haskell haskellPackages rhine-all;
-      });
+      # We re-export the entire nixpkgs package set with our overlay.
+      # Usage examples:
+      # - nix build .#haskellPackages.rhine
+      # - nix build .#haskell.packages.ghc98.rhine
+      # - nix build .#rhine-sdist
+      legacyPackages = forAllPlatforms (system: pkgs: pkgs);
 
-      # Usage: nix develop .#ghc98
+      # Usage: nix develop (will use the default GHC)
+      # Alternatively, specify the GHC: nix develop .#ghc98
       devShells = forAllPlatforms (systems: pkgs: mapAttrs
         (_: hp: hp.shellFor {
           packages = ps: map (pname: ps.${pname}) pnames;
--- a/flake.nix
+++ b/flake.nix
@@ -22,81 +22,122 @@
 
       # The names of all Haskell packages in this repository, defined as all the directories with *.cabal files in them.
       pnames = map (path: baseNameOf (dirOf path)) (lib.fileset.toList (lib.fileset.fileFilter (file: file.hasExt "cabal") ./.));
+
+      # All GHC versions that this project is tested with.
+      # To be kept in sync with the `tested-with:` section in rhine.cabal.
+      # To do: Automated check whether this is the same as what get-tested returns.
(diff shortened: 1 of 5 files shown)


turion:
Thanks a lot! This is awesome!

flake.nix
```diff
@@ -13,56 +13,76 @@
 
   inputs = {
     nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable-small";
-    flake-utils.url = "github:numtide/flake-utils";
-
-    flake-compat = {
-      url = "github:edolstra/flake-compat";
+    monad-schedule = {
+      url = "github:turion/monad-schedule";
```
turion, line 17: I don't quite understand why we would want to pull in `master` from this repo. My maintenance commitment is only to the Hackage published versions within the bounds in `rhine.cabal`.

flake.nix
```diff
@@ -13,56 +13,76 @@
...
+            prev.haskell.packageOverrides
+            rhinePackages
+            (hfinal: hprev: {
+              monad-bayes = markUnbroken hprev.monad-bayes;
+              monad-schedule = hfinal.callCabal2nix "monad-schedule" inputs.monad-schedule { };
+              time-domain = hfinal.callCabal2nix "time-domain" inputs.time-domain { };
+            })
+            (hfinal: hprev: lib.optionalAttrs (lib.versionOlder hprev.ghc.version "9.4") {
+              time-domain = doJailbreak hprev.time-domain;
+            })
```
turion, line 56: I'd rather use callHackageDirect here than refer to github. After all, I'm maintaining a build with Debian + cabal in another CI job as well, so overrides will only be necessary whenever a Hackage version is not on stackage/nixpkgs.

If `callHackageDirect` is a poor choice for some other reason, we could still use Github and specific tags corresponding to versions. But I don't want to use `master` for these dependencies because I might push a breaking version.

turion:
The test failure may simply be the Github MacOS runner. For some reason, some concurrency and timing stuff from monad-schedule doesn't work on MacOS so well.

flake.nix
ners: I went with `callHackageDirect`, since that's what we had before.

My original thinking was that we might not want to keep updating the flake with new hashes every time these dependencies update. Ideally that's something that flake inputs can automate.

flake.nix
```diff
@@ -13,56 +13,79 @@
...
+      lib = inputs.nixpkgs.lib;
+      pnames = [
+        "automaton"
+        "rhine"
+        "rhine-bayes"
+        "rhine-examples"
+        "rhine-gloss"
+        "rhine-terminal"
+      ];
+      supportedGhcs = [ "ghc92" "ghc94" "ghc96" "ghc98" ];
```
turion, line 30: Is there any way to easily build these packages for just one specific version of GHC?

flake.nix
```diff
@@ -13,56 +13,79 @@
...
+      hpsFor = pkgs:
+        lib.filterAttrs (name: _: elem name supportedGhcs) pkgs.haskell.packages
```
turion, line 32: I think I don't really understand this. Isn't the aim to build this package for all supported GHCs?

flake.nix
ners: Yes. We export `haskell` and `haskellPackages`, so you can do something like:

```
$ nix build .#haskell.packages.ghc98.rhine
```

This is not limited to `supportedGhcs` either, it can be built with any GHC version in nixpkgs.

flake.nix
ners: `hpsFor` converts `pkgs` into an attribute set of `{ ${ghcName} = pkgs.haskell.packages.${ghcName}; }` for every supported `ghcName`.

This allows us to easily map over this attribute set e.g. when creating `rhine-all` or per-GHC devshells.

turion: Makes sense! One detail: This silently drops those attribute names that are listed in `supportedGhcs` but do not occur in `pkgs.haskell.packages`? I think it would be better if there was a loud error when e.g. a ghc version is misspelled, or not maintained in nixpkgs anymore. Let me try and adapt that.

flake.nix
turion: Great, I tried it, works :)

flake.nix
```diff
@@ -13,56 +13,112 @@
...
+      lib = inputs.nixpkgs.lib;
+      pnames = [
+        "automaton"
+        "rhine"
+        "rhine-bayes"
+        "rhine-examples"
+        "rhine-gloss"
+        "rhine-terminal"
+      ];
+      supportedGhcs = [ "ghc92" "ghc94" "ghc96" "ghc98" ];
```
turion, line 30: Challenge to myself: A CI job that compares this list to `get-tested`.

flake.nix
```diff
@@ -13,56 +13,112 @@
...
+      });
 
-    name = "rhine";
-    packageNames = [ "automaton" "rhine-gloss" "rhine-terminal" "rhine-examples" "rhine-bayes" ];
-  };
+      legacyPackages = forAllPlatforms (system: pkgs: {
+        inherit (pkgs) haskell haskellPackages rhine-all;
+      });
+
+      devShells = forAllPlatforms (systems: pkgs: mapAttrs
```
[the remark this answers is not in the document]
ners: Try: `nix develop .#ghc98`

If using nix-direnv you can put this in the `.envrc`:
```
use flake .#ghc98
```

flake.nix
ners: For that I'd suggest replacing `filterAttrs` with `genAttrs`.

flake.nix
```diff
@@ -13,56 +13,112 @@
...
+            paths = lib.mapCartesianProduct
+              ({ hp, pname }: hp.${pname})
+              { hp = attrValues hps; pname = pnames; };
+            pathsToLink = [ "/lib" ];
+          };
+          rhine-docs = prev.buildEnv {
+            name = "rhine-docs";
+            paths = map (pname: prev.haskell.lib.documentationTarball hps.default.${pname}) libPnames;
+          };
+          rhine-sdist = prev.buildEnv {
```
turion, line 84: Can I access these `rhine-*` derivations directly in any way? Or are they just prerequisites for `rhine-all`?

flake.nix
```diff
@@ -13,56 +13,112 @@
...
+              ln -s ${final.rhine-docs} $out/docs
+              ln -s ${final.rhine-sdist} $out/sdist
+            '';
+          };
+        };
+      forAllPlatforms = f:
+        mapAttrs (system: pkgs: f system (pkgs.extend overlay)) inputs.nixpkgs.legacyPackages;
+    in
+    {
+      overlays.default = overlay;
```
turion, line 101: How would you use this flake in a downstream rhine project?

flake.nix
```diff
@@ -13,56 +13,112 @@
...
+            name = "rhine-sdist";
+            paths = map (pname: prev.haskell.lib.sdistTarball hps.default.${pname}) libPnames;
+          };
+          rhine-all = prev.symlinkJoin {
+            name = "rhine-all";
+            paths = with final; [ rhine-bin rhine-lib ];
+            postBuild = ''
+              ln -s ${final.rhine-docs} $out/docs
+              ln -s ${final.rhine-sdist} $out/sdist
+            '';
```
turion, line 94: Wouldn't this be easier to achieve with `linkFarm`?

flake.nix
```diff
@@ -13,56 +13,112 @@
...
+              (hfinal: hprev: lib.optionalAttrs (lib.versionOlder hprev.ghc.version "9.4") {
+                time-domain = doJailbreak hprev.time-domain;
+              })
+            ];
+          };
+          rhine-bin = prev.buildEnv {
+            name = "rhine-bin";
+            paths = map (pname: hps.default.${pname}) pnames;
+            pathsToLink = [ "/bin" ];
+          };
```
turion, line 72: I'm not sure what this is for. The executables should be built by `rhine-lib` as well, right?

turion: Ah, but in `/lib` there is a directory structure that disambiguates the ghc version. Since this doesn't exist in `/bin` we need to chose a GHC version here, hence the different treatment (if we want the executables as a build output).

turion: Let's leave it like this for now. I'm wondering whether there is value in having a structure `/bin/ghc98/executable1, /bin/ghc96/executable2`, and so on.

turion:
I added a few commits to document everything. I needed this in order to fully understand what's going on. I believe I have it now. In case you can spot a blatant misunderstanding, feel free to call it out :)

flake.nix
ners: Here's an example project where I already use it: https://github.com/ners/rhine-discord/blob/d6db1feca7efe68eadf6b8e0f84896383eab5e58/flake.nix#L32

flake.nix
ners: Good point. I've exposed them by simply making `legacyPackages` point to `pkgs`. This way we can access everything in our overlay.

flake.nix
turion: #342

flake.nix
turion: Sorry, I had to force-push over that change. Can you apply it again?

flake.nix
turion: I'll do something like that in https://github.com/turion/rhine-koans/issues/53, thanks!

turion:
Fantastic! Again, thanks a lot, this makes rhine development a lot more productive. And I've learned a lot along the way.

PR #346, opened 2024-08-06
Fix naming of initial and final

turion:
For some reason, I misnamed initial and final algebras, and direct recursion. This PR fixes it.


turion:
Thinking about it once again I notice that this is not final nor initial encoding, but simply coalgebraic encoding.

Issue #365, opened 2024-10-21
Parallel composition of FixedStep clocks evaluates in unexpected order


DISTEL100:
There seems to be a problem with the parallel clock when composing two pure clocks in `ScheduleT`.  The merged output is not ordered by the timestamps. 

The following is a minimal example to reproduce the issue:
```haskell

f300 :: (MonadIO m) => Rhine (ScheduleT Integer m) (FixedStep 300) a Integer
f300 =  absoluteS @@ FixedStep

f500 :: (MonadIO m) => Rhine (ScheduleT Integer m) (FixedStep 500) a Integer
f500 =  absoluteS @@ FixedStep

f :: IO ()
f = runScheduleIO $ flow $ (f300 +@+ f500) @>-^ arrM (liftIO . print)
```
The output is:
```bash
ghci> f
Left 300
Left 600
Right 500
Left 900
Left 1200
Right 1000
Left 1500
Left 1800
Right 1500
Left 2100
Left 2400
Right 2000
Left 2700
Right 2500
Left 3000
Left 3300
^CInterrupted.

```

I would expect it to be
```bash
ghci> f
Left 300
Right 500
Left 600
Left 900
Right 1000
Left 1200
Left 1500
Right 1500
Left 1800
Right 2000
Left 2100
Left 2400
Right 2500
Left 2700
Left 3000
Left 3300
^CInterrupted.

```

turion:
Thanks for finding this! I can sort of reproduce, but I'm getting a different (yet still incorrect) scheduling:

```
ghci> runScheduleIO $ flow $ (f300 +@+ f500) @>-^ arrM (liftIO . print)
Left 300
Right 500
Left 600
Left 900
Left 1200
Right 1000
Left 1500
Left 1800
Left 2100
Right 1500
Left 2400
Right 2000
Left 2700
Left 3000
^CInterrupted.
```

turion:
I have an inkling that this might be an upstream issue in https://github.com/turion/monad-schedule. Let me investigate.

turion:
Which versions of `rhine` and `automaton` are you using?

DISTEL100:
i have rhine ==1.4.0.1 and monad-schedule 0.2.0.1 and i just ran it again and had the same output like you 
> Thanks for finding this! I can sort of reproduce, but I'm getting a different (yet still incorrect) scheduling:
> 
> ```
[...]

PR #368, opened 2024-10-27
Dev regression #365

turion:
Regression test for #365


turion:
As expected, this fails before https://github.com/turion/rhine/pull/343 is merged.

turion:
Even after merging #343, this is still flaky.

turion:
It seems ok on GHC 9.6, but flaky on GHC 9.8!

turion:
But `nix build .#haskell.packages.ghc96.rhine` is also flaky on my machine.

turion:
Now that the issue is solved, this test could be rebased and merged.

PR #371, opened 2024-11-14
Fix monad-schedule flake input

ners:
This prevents the [1000 instances of nixpkgs](https://zimbatm.com/notes/1000-instances-of-nixpkgs) problem.

I also drive-by re-enable flake checks. :angel:


turion:
Thanks a lot!

(I really need to do somethin about that flaky test suite of monad-schedule)

PR #373, opened 2024-11-16
Build monad-schedule without parallelism in CI

ners:
`monad-schedule` has a test suite that is real-time-sensitive, and thus building it for multiple GHCs with high parallelism makes the tests flaky.

This PR ensures that `monad-schedule` is built with `-j1` in CI.


.github/workflows/ci.yml
```diff
@@ -132,7 +132,10 @@ jobs:
       - name: Flake check
         run: nix flake check
       - name: Build all packages
-        run: nix build --accept-flake-config
+        run: |
+          # Pre-build monad-schedule without parallelism to avoid timing issues under load
+          nix build --accept-flake-config --max-jobs 1 .#monad-schedule-lib
+          nix build --accept-flake-config
```
turion, line 138: Preferably these were 2 jobs
```suggestion
        run: |
          # Pre-build monad-schedule without parallelism to avoid timing issues under load
          nix build --accept-flake-config --max-jobs 1 .#monad-schedule-lib
          nix build --accept-flake-config
```

PR #424, opened 2026-04-12
time-domain: change Diff UTCTime to NominalDiffTime

ners:
The canonical `Diff UTCTime` is `NominalDiffTime`, not `Double`.

This exposes a number of assumptions / unit confusions throughout the codebase; e.g. `integralFrom` multiplies a vector by a time difference instead of a scalar.

We introduce a new utility class called `Measured` which explicitly generalises this assumption.

An even more principled approach would be to split out the integration stuff into a separate library and use a coherent unit library such as `dimensional`, e.g. `rhine-dimensional`.


turion:
> The canonical Diff UTCTime is NominalDiffTime, not Double.

I agree on that part, this is a really good point in principle. But could we also just change all the corresponding type signatures to use `NominalDiffTime` instead? Or even make them more polymorphic so that they would expect both `NominalDiffTime` or `Double`?

For `AudioClock`, I'm thinking that actually `Double` is a terrible idea, but the correct solution would be some newtype on `Integer` that counts the sample.

In general I'm not completely on board with the `Measured` class. To be honest it feels like an ad hoc solution to something that should already have a good solution. Actually, a [`Rescaling](https://hackage-content.haskell.org/package/rhine-1.6/docs/FRP-Rhine-Clock.html#t:Rescaling) is the concept that is intended for this situation, and I'm wondering why it's not working here. Can you tell me about the situation where you stumbled into a problem? Maybe a code example?

PR #430, opened 2026-05-10
time-domain: add epoch and zero duration

ners:
When writing code that uses `TimeDomain time`, I often find myself having to add additional constraints on `time` or `Diff time` in order to be able to express "initial accumulator value" or "duration of an empty sequence of timestamps".

One such trick is to declare `Num time` and use `0`, or declare `Monoid time` and use `mempty`.

Such tricks are not satisfying.

This PR makes two assertions, that are hopefully not too controversial:
- Every time domain has a well-defined, canonical origin / reference value
- Every set of time differences has an additive identity for `add` and `difference`

### DIFF
--- a/time-domain/src/Data/TimeDomain.hs
+++ b/time-domain/src/Data/TimeDomain.hs
@@ -55,10 +55,15 @@ class (TimeDifference (Diff time)) => TimeDomain time where
 
 Expected laws:
 
+* @dt `add` 'zero' = dt@ and @dt `difference` 'zero' = dt@
 * `add` is commutative and associative
 * @(dt1 `difference` dt2) `add` dt2 = dt1@
 -}
 class TimeDifference d where
+  -- | The zero duration, i.e. no time has passed.
+  --   Additive identity for 'difference' and 'add'.
+  zero :: d
+
   -- | Calculate the difference between two durations,
   --   compatibly with 'diffTime'.
   difference :: d -> d -> d
@@ -74,6 +79,7 @@ instance TimeDomain UTCTime where
   addTime = flip $ addUTCTime . realToFrac
 
 instance TimeDifference Double where
+  zero = 0
   difference = (-)
   add = (+)
 
@@ -84,6 +90,7 @@ instance TimeDomain Double where
   addTime = (+)
 
 instance TimeDifference Float where
+  zero = 0
   difference = (-)
   add = (+)
 
@@ -94,6 +101,7 @@ instance TimeDomain Float where
   addTime = (+)
 
 instance TimeDifference Integer where
+  zero = 0
   difference = (-)
   add = (+)
 
@@ -104,6 +112,7 @@ instance TimeDomain Integer where
   addTime = (+)
 
 instance TimeDifference () where
+  zero = ()
   difference _ _ = ()
   add _ _ = ()
 
@@ -118,6 +127,7 @@ newtype NumTimeDomain a = NumTimeDomain {fromNumTimeDomain :: a}
   deriving (Num)
 
 instance (Num a) => TimeDifference (NumTimeDomain a) where
+  zero = 0
   difference = (-)
   add = (+)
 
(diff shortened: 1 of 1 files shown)


turion:
This is a great idea! For coordination, I'm about to push some changes to make TimeDifference a Monoid, which would make null obsolete (or at least give it a default implementation). Let me briefly think about what to merge first.

time-domain/src/Data/TimeDomain.hs
```diff
@@ -32,6 +33,14 @@ class (TimeDifference (Diff time)) => TimeDomain time where
...
   type Diff time
 
+  -- | The origin of the time domain.
+  --
+  --   All timestamps can be understood as durations relative to this reference point.
+  --
+  --   For example, 'UTCTime' uses the Unix epoch (1970-01-01 00:00:00 UTC),
+  --   while numeric instances use @0@.
+  epoch :: time
+
```
turion, line 43: Would you also be fine to move this to a sub-typeclass `TimeDomainEpoch` or `TimeEpoch` or `EpochDomain`? I can't really think of the method being necessary in any Rhine internals, and there is some potential of confusion with `initialTime`.

ners: I‘d invert this question to ask: is there a time domain in which a concept of epoch does not exist? I think it‘s a meaningful addition to all time domains.

ners: > there is some potential of confusion with `initialTime`

A Rhine clock can already be instantiated with `0` or UNIX epoch or whatever other epoch one wishes. If that is semantically appropriate for a given clock, then I don‘t see a problem.

ners:
Neat, making all `TimeDifference`s a `Monoid` would automatically make them eligible for `MonadChangeset` as well.

time-domain/src/Data/TimeDomain.hs
turion: > I‘d invert this question to ask: is there a time domain in which a concept of epoch does not exist? I think it‘s a meaningful addition to all time domains.

I don't think it's a necessary part of the definition of a time domain. Arguably the UNIX epoch is only an implementation detail. Why not take the year 0 as epoch instead?

My point is rather pragmatic: `TimeDomain` is required in many functions, but a functionality like `epoch` isn't needed in any of them.

What's a use case for `epoch`? Maybe I don't understand what you need it for. All use cases where I feel like needing a reference time are usually covered either by the initial time which is only known at runtime, or `Nothing` (where no timestamp so far is an error case).

ners: I'm using `time-domain` in libraries that do not depend on `rhine`, so the concept of an `initialTime` often does not exist. So far I've either pushed a reference time as a parameter to various functions, or used `mempty` as a poor man's proxy (it happens to work fine for `Dimensional Time` at least).

ners: > I don't think it's a necessary part of the definition of a time domain.

What is a timestamp if not an offset from a reference point?

turion: Fair enough, so would a separate type class work there as well, or would it be a hindrance?

ners: It would work, in the same way having a `Monoid` instance currently works.

I can't think of a single time domain, in Rhine or in my own codebases, that does not have a natural zero.

I suppose I'm proposing to redefine `TimeDomain` from an affine space to a vector space. :)

ners:
If you wish to keep `TimeDomain` an affine space, then `Monoid` is already a good enough abstraction that extends a time domain to a vector space, so there is no value to be gained in a separate epoch type class.

With the planned `Monoid` interface for `TimeDifference`, both concerns of this PR are addressed.

turion:
Ok, if you're fine with it I'll still cherry pick your commit adding the method for `TimeDifference`.

PR #442, opened 2026-06-14
Fix 441 performance first

turion:
Running `cabal bench benchmark --benchmark-options="-m pattern flow"` shows that the performance on the sum rhine flow benchmark improves by >100x.

PR #443, opened 2026-06-14
Simplify clock class definition

turion:
* [ ] #442


rhine/src/FRP/Rhine/Clock.hs
```diff
@@ -67,11 +57,11 @@ class (TimeDomain (Time cl)) => Clock m cl where
...
-  initClock ::
-    -- | The clock value, containing e.g. settings or device parameters
-    cl ->
-    -- | The stream of time stamps, and the initial time
-    RunningClockInit m (Time cl) (Tag cl)
+  runClock ::
+    -- | The stream of time stamps and tags,
+    --   which may depend on the clock value, containing e.g. settings or device parameters.
+    --   It may produce side effects in @m@ to read the current time or wait for the next tick.
+    Automaton m cl (Time cl, Tag cl)
```
turion, line 65: @ners I found a clock definition that I find sensible that seems to have little problems. Do you think it is possible to adapt all your clock definitions to that one? If not, let's have a look how we can translate them. It's important to me that the new definition doesn't lose any expressivity.

